PR #5, opened 2015-01-02
Type safety in statement parsing

cchantep:
- Avoid clash of reserved `%`: http://stackoverflow.com/questions/27697588/possible-bug-in-jdbc
- Make parsed statement more type safe
- No longer have to 'search' for placeholder to be prepared as SQL (as tokenized statement structure allows direct access to): runtime parsing (`SQL("...")`) has better performance, same performance for string interpolation (no more load at runtime due to this more complex/type safe tokenization, thanks to the macro helper).

### DIFF
--- a/src/main/scala/anorm/Anorm.scala
+++ b/src/main/scala/anorm/Anorm.scala
@@ -309,36 +309,11 @@ object Sql { // TODO: Rename to SQL
     case _ => ps
   }
 
-  /**
-   * Rewrites next format placeholder (%s) in statement, with fragment using
-   * [[java.sql.PreparedStatement]] syntax (with one or more '?').
-   *
-   * @param statement SQL statement (with %s placeholders)
-   * @param frag Statement fragment
-   * @return Some rewrited statement, or None if there no available placeholder
-   *
-   * {{{
-   * Sql.rewrite("SELECT * FROM Test WHERE cat IN (%s)", "?, ?")
-   * // Some("SELECT * FROM Test WHERE cat IN (?, ?)")
-   * }}}
-   */
-  private[anorm] def rewrite(stmt: String, frag: String): Option[String] = {
-    val idx = stmt.indexOf("%s")
-
-    if (idx == -1) None
-    else {
-      val parts = stmt.splitAt(idx)
-      Some(parts._1 + frag + parts._2.drop(2))
-    }
-  }
-
   @annotation.tailrec
-  private[anorm] def prepareQuery(sql: String, i: Int, ps: Seq[ParameterValue], vs: Seq[(Int, ParameterValue)]): (String, Seq[(Int, ParameterValue)]) = {
-    ps.headOption match {
-      case Some(p) =>
-        val st: (String, Int) = p.toSql(sql, i)
-        prepareQuery(st._1, st._2, ps.tail, vs :+ (i -> p))
-      case _ => (sql, vs)
-    }
+  private[anorm] def prepareQuery(stmt: TokenizedStatement, i: Int, ps: Seq[ParameterValue], vs: Seq[(Int, ParameterValue)]): (TokenizedStatement, Seq[(Int, ParameterValue)]) = ps.headOption match {
+    case Some(p) =>
+      val st: (TokenizedStatement, Int) = p.toSql(stmt, i)
+      prepareQuery(st._1, st._2, ps.tail, vs :+ (i -> p))
+    case _ => (stmt, vs)
   }
 }
--- a/src/main/scala/anorm/BatchSql.scala
+++ b/src/main/scala/anorm/BatchSql.scala
@@ -116,10 +116,12 @@ sealed trait BatchSql {
   private def fill(con: Connection, statement: PreparedStatement, getGeneratedKeys: Boolean = false, pm: Seq[Map[String, ParameterValue]]): PreparedStatement =
     (statement, pm.headOption) match {
       case (null, Some(ps)) => { // First
-        val st: (String, Seq[(Int, ParameterValue)]) = Sql.prepareQuery(
-          sql.statement, 0, sql.paramsInitialOrder.map(ps), Nil)
+        val st: (TokenizedStatement, Seq[(Int, ParameterValue)]) =
+          Sql.prepareQuery(sql.stmt, 0,
+            sql.paramsInitialOrder.map(ps), Nil)
 
-        val stmt = if (getGeneratedKeys) con.prepareStatement(st._1, java.sql.Statement.RETURN_GENERATED_KEYS) else con.prepareStatement(st._1)
+        val psql = TokenizedStatement.toSql(st._1).get // TODO: Make it safe
+        val stmt = if (getGeneratedKeys) con.prepareStatement(psql, java.sql.Statement.RETURN_GENERATED_KEYS) else con.prepareStatement(psql)
 
         sql.timeout.foreach(stmt.setQueryTimeout(_))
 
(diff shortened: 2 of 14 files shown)


src/main/scala/anorm/Anorm.scala
```diff
@@ -309,36 +309,11 @@ object Sql { // TODO: Rename to SQL
     case _ => ps
   }
 
-  /**
```
cchantep: Refactored and moved to `TokenizedStatement` object

src/main/scala/anorm/Anorm.scala
```diff
@@ -309,36 +309,11 @@ object Sql { // TODO: Rename to SQL
...
-  private[anorm] def prepareQuery(sql: String, i: Int, ps: Seq[ParameterValue], vs: Seq[(Int, ParameterValue)]): (String, Seq[(Int, ParameterValue)]) = {
-    ps.headOption match {
-      case Some(p) =>
-        val st: (String, Int) = p.toSql(sql, i)
-        prepareQuery(st._1, st._2, ps.tail, vs :+ (i -> p))
-      case _ => (sql, vs)
-    }
+  private[anorm] def prepareQuery(stmt: TokenizedStatement, i: Int, ps: Seq[ParameterValue], vs: Seq[(Int, ParameterValue)]): (TokenizedStatement, Seq[(Int, ParameterValue)]) = ps.headOption match {
+    case Some(p) =>
+      val st: (TokenizedStatement, Int) = p.toSql(stmt, i)
```
cchantep: Use `TokenizedStatement` instead of internal string

src/main/scala/anorm/BatchSql.scala
```diff
@@ -116,10 +116,12 @@ sealed trait BatchSql {
   private def fill(con: Connection, statement: PreparedStatement, getGeneratedKeys: Boolean = false, pm: Seq[Map[String, ParameterValue]]): PreparedStatement =
     (statement, pm.headOption) match {
       case (null, Some(ps)) => { // First
-        val st: (String, Seq[(Int, ParameterValue)]) = Sql.prepareQuery(
-          sql.statement, 0, sql.paramsInitialOrder.map(ps), Nil)
+        val st: (TokenizedStatement, Seq[(Int, ParameterValue)]) =
```
cchantep: Use `TokenizedStatement` instead of internal string

src/main/scala/anorm/SimpleSql.scala
```diff
@@ -54,10 +54,11 @@ case class SimpleSql[T](sql: SqlQuery, params: Map[String, ParameterValue], defa
 
   @deprecated(message = "Use [[preparedStatement]]", since = "2.3.6")
   def getFilledStatement(connection: Connection, getGeneratedKeys: Boolean = false) = {
-    val st: (String, Seq[(Int, ParameterValue)]) = Sql.prepareQuery(
+    val st: (TokenizedStatement, Seq[(Int, ParameterValue)]) = Sql.prepareQuery(
```
cchantep: Use `TokenizedStatement` instead of internal string

src/main/scala/anorm/ParameterValue.scala
```diff
@@ -14,7 +14,8 @@ sealed trait ParameterValue {
    * @return Update statement with '?' placeholder(s) for parameter,
    * with offset for next parameter
    */
-  def toSql(stmt: String, offset: Int): (String, Int)
+  def toSql(stmt: TokenizedStatement, offset: Int): (TokenizedStatement, Int)
```
cchantep: Use `TokenizedStatement` instead of internal string

src/main/scala/anorm/SqlQuery.scala
```diff
@@ -45,14 +46,14 @@ sealed trait SqlQuery {
...
-  def asBatch[T]: BatchSql = BatchSql(this.query, Nil)
+  def asBatch[T]: BatchSql = BatchSql.Checked(this, Nil)
 
   private def defaultParser: RowParser[Row] = RowParser(Success(_))
 
   // TODO: Make it private
   @deprecated(
     "Use anorm.SQL(…) as SqlQuery should not be directly created", "2.3.2")
-  def copy(statement: String = this.statement, paramsInitialOrder: List[String] = this.paramsInitialOrder, timeout: Option[Int] = this.timeout) = SqlQuery(statement, paramsInitialOrder, timeout)
+  def copy(statement: TokenizedStatement = this.statement, paramsInitialOrder: List[String] = this.paramsInitialOrder, timeout: Option[Int] = this.timeout) = SqlQuery(statement, paramsInitialOrder, timeout)
```
cchantep: Use `TokenizedStatement` instead of internal string

src/main/scala/anorm/SqlQuery.scala
```diff
@@ -6,7 +6,7 @@ import resource.ManagedResource
 /** Initial SQL query, without parameter values. */
 sealed trait SqlQuery {
   /** SQL statement */
-  def statement: String
```
cchantep: Not backward compatible due to this one

tokenizer/src/main/scala/anorm/TokenizedStatement.scala
```diff
@@ -0,0 +1,139 @@
...
+  import scala.language.experimental.macros
+  
+  /** Returns empty tokenized statement. */
+  lazy val empty = TokenizedStatement(Nil, Nil)
+
+  /** String interpolation to tokenize statement. */
+  def stringInterpolation[T](parts: Seq[String], params: Seq[T with Show]): (TokenizedStatement, Map[String, T]) = macro tokenizeImpl[T]
+
+  /** Tokenization macro */
+  def tokenizeImpl[T: c.WeakTypeTag](c: Context)(parts: c.Expr[Seq[String]], params: c.Expr[Seq[T with Show]]): c.Expr[(TokenizedStatement, Map[String, T])] =
```
cchantep: Allow to tokenize at compile time for Anorm string interpolation (no extra load at runtime).

tokenizer/build.sbt
```diff
@@ -0,0 +1,5 @@
+name := "anorm-tokenizer"
```
cchantep: As sub-project to compile macro before main Anorm code

cchantep:
This one should be ready for merge now.

cchantep:
@jroper Could have a look at this one?

tokenizer/build.sbt
jroper: I think some of the settings (the publish/maven settings) in the main `build.sbt` need to be in here too.  They could be added to the `project/Common.scala` plugin, but that means it will also be added to the docs project, which I'm not sure if we want or not.

jroper:
Hey @cchantep, sorry for taking a while, I've been putting off reviewing this because I've been a bit busy.

So from what I gather, from an end users perspective, nothing really changes in the API (unless they're using a few low level API methods that aren't really supposed to be used).  It's just that you now get better performance, and compile time errors if statement parsing fails.

By the way, what does error reporting look like?  The biggest issues I've seen with macros are that errors the user ends up seeing are cryptic, do the errors in this case make sense?  I'm wondering if we can test the error handling and error messages at all, possibly using Twitter Eval, or maybe even just firing up the Scala compiler manually.

cchantep:
The macro is mainly wrapping standard interpolation (compile errors handled by that standard macro for me), with specific behaviour tested in `SqlStatementSpec` (about `#$`).

Moreover performance, this fix trick of Anorm placeholders and their preparation it won't introduce reserved characters (`%s`) that clash if in the raw statement (see the SO link).

tokenizer/build.sbt
cchantep: SBT files have been refactored

cchantep:
Thanks @jroper

PR #7, opened 2015-01-15
Use different name for Success in tokenizer project

pvlugter:
This is to avoid `anorm.Success` being used when building the combined scaladoc in omnidoc.

### DIFF
--- a/tokenizer/src/main/scala/anorm/TokenizedStatement.scala
+++ b/tokenizer/src/main/scala/anorm/TokenizedStatement.scala
@@ -1,6 +1,6 @@
 package anorm
 
-import scala.util.{ Failure, Try, Success }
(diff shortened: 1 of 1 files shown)


cchantep:
For my information, package is not taken in account in omnidoc?

pvlugter:
Omnidoc builds a combined scaladoc, running scaladoc over all the play module sources. In this case it means that the tokenizer project is compiled alongside the anorm project, and then `scala.util.Success` is hidden by `anorm.Success`.

We could add a combined scaladoc to the anorm build, which would fail in the same way. But this is the first time we've had any issues, and normally there shouldn't be any need to workaround like this. And we want to eventually replace omnidoc with properly modular docs.

PR #9, opened 2015-01-24
Missing documentation about passing optional value as parameter

cchantep:
See http://stackoverflow.com/questions/28096650/null-parameter-in-anorm-queries


cchantep:
Should be ok

PR #10, opened 2015-01-24
Support Java8 Instant

cchantep:
See http://stackoverflow.com/questions/28031715/anorm-with-java-8-date-time-api


docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md
```diff
@@ -911,10 +911,10 @@ Vector                    | Multi-value, with `T` mapping for each element
 JVM                  | JDBC
 ---------------------|-----------
 DateTime<sup>1</sup> | Timestamp
-Instant<sup>2</sup>  | Timstamp
+Instant<sup>2</sup>  | Timestamp
```
cchantep: Fix typo

jroper:
Should we put this in a separate project?  ie, an `anorm-java8-compat` project?  That way, the core of anorm can still be compiled and published for Java 1.6, while this java8 only library can be an optional addition for those on Java 8?

cchantep:
It can. Current PR already exclude Java8 code if <= 1.7

jroper:
Yes, but that's only when building anorm that it gets included/excluded.  For end users that just add an anorm dependency, we'll have build it as a Java 1.6 library, and push that to maven central.  They won't get the Java 8 support - the only way they could get it is if they built anorm themselves locally.

build.sbt
```diff
@@ -1,39 +1,48 @@
+import scala.util.Properties.isJavaAtLeast
+
 lazy val tokenizer = project
   .in(file("tokenizer"))
   .enablePlugins(Omnidoc, Publish)
 
-lazy val anorm = project
+lazy val core = project
```
cchantep: Separate and gather core project

build.sbt
```diff
@@ -1,39 +1,48 @@
...
+    ).map("org.specs2" %% _ % "2.4.9" % Test)
+  }).dependsOn(tokenizer)
+
+lazy val java8 = project
+  .in(file("java8"))
+  .dependsOn(core)
+  .enablePlugins(Omnidoc, Publish)
+  .settings(javacOptions := Seq("-source", "1.8", "-target", "1.8"))
+
+lazy val anorm = project.aggregate(tokenizer, core) configure { p =>
```
cchantep: Conditional project aggregation

cchantep:
@jroper should be ok

build.sbt
jroper: I think this changes the artifact id of the project (sbt uses a macro that finds the name of the val), we can override it specifically by setting `projectId := "anorm"` or something like that, but it may be easier to leave this as `anorm`, and rename the `anorm` project down lower to `root`.  Also, maybe it's time to restructure the project a little, ie put the core project in a directory called `core`.

build.sbt
```diff
@@ -1,39 +1,49 @@
...
-    Seq("org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.1")
-  case _ => Nil
-})
-
-libraryDependencies ++= Seq(
-  "specs2-core",
-  "specs2-junit",
-  "specs2-mock"
-).map("org.specs2" %% _ % "2.4.9" % Test)
+name := "root"
```
cchantep: Aggregation root

build.sbt
```diff
@@ -1,39 +1,49 @@
...
+    ).map("org.specs2" %% _ % "2.4.9" % Test)
+  }).dependsOn(tokenizer)
+
+lazy val java8 = project
+  .in(file("java8"))
+  .dependsOn(anorm)
+  .enablePlugins(Omnidoc, Publish)
+  .settings(javacOptions := Seq("-source", "1.8", "-target", "1.8"))
+
+lazy val root = Project(id = "anorm-parent", base = file(".")).
```
cchantep: Aggregation descriptor

build.sbt
```diff
@@ -1,39 +1,49 @@
+import scala.util.Properties.isJavaAtLeast
+
 lazy val tokenizer = project
   .in(file("tokenizer"))
   .enablePlugins(Omnidoc, Publish)
 
 lazy val anorm = project
```
cchantep: Descriptor of main sub-project whose sources are moved in subdir `core`.

PR #15, opened 2015-02-12
BatchSql, NullPointerException with empty or missing parameters

cchantep:
Anorm from play 2.3.7
Scala 2.11

    DB.withConnection { implicit c =>
        BatchSql(SQL("update sometable set something = 1600 where id = 2")).execute()
    }
explicitly: (even though Nil is default value for parameter)

    DB.withConnection { implicit c =>
        BatchSql(SQL("update sometable set something = 1600 where id = 2"), Nil).execute()
    }

My use case: 
`BatchSql(SQL("somesql"), aListThatMayOrMayNotBeEmpty))`

Expected:
If list is non-empty, run sql (in batch) with provided parameters, if list is empty, do nothing.

What happened:
`Execution exception[[NullPointerException: null]]`
Not very descriptive.

I spent way too much time on this just to come to the conclusion that parameters could not be an empty list (even though according to doc/source BatchSql defaults parameter "ps" to Nil)

### DIFF
--- a/core/src/test/scala/anorm/H2Database.scala
+++ b/core/src/test/scala/anorm/H2Database.scala
@@ -5,11 +5,10 @@ import scala.util.Random
 
 trait H2Database {
 
-  def withH2Database[R](block: Connection => R) = {
-    val dbName = "test" + Random.alphanumeric.take(6).mkString("")
+  def withH2Database[R](block: Connection => R): R = {
+    val url = "jdbc:h2:mem:test" + Random.alphanumeric.take(6).mkString("")
+    val connection = DriverManager.getConnection(url, "sa", "")
 
-    Class.forName("org.h2.Driver")
-    val connection = DriverManager.getConnection("jdbc:h2:mem:" + dbName, "sa", "")
     try {
       block(connection)
     } finally {
--- a/core/src/main/scala/anorm/BatchSql.scala
+++ b/core/src/main/scala/anorm/BatchSql.scala
@@ -113,28 +113,35 @@ sealed trait BatchSql {
   }
 
   @annotation.tailrec
-  private def fill(con: Connection, statement: PreparedStatement, getGeneratedKeys: Boolean = false, pm: Seq[Map[String, ParameterValue]]): PreparedStatement =
-    (statement, pm.headOption) match {
-      case (null, Some(ps)) => { // First
-        val st: (TokenizedStatement, Seq[(Int, ParameterValue)]) =
-          Sql.prepareQuery(sql.stmt, 0,
-            sql.paramsInitialOrder.map(ps), Nil)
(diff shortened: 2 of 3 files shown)


cchantep:
No longer failing with exception with this PR, but passing `Nil` or `List.empty` will lead to execute a single statement w/o parameter, which is meaningless for `BatchSql`.

optama:
@cchantep How would you suggest one should combine using BatchSql with a list that might or might not be empty? 

At the moment I'm doing
```
if (parameters.nonEmpty)
    BatchSql(SQL("update sometable set somevalue = {value} where id = {id}"), parameters.map(p => List[NamedParameter]("value" -> p.value, "id" -> p.id))).execute()
```

Is this as intended - as opposed to 

    BatchSql(SQL("update sometable set somevalue = {value} where id = {id}"), parameters.map(p => List[NamedParameter]("value" -> p.value, "id" -> p.id))).execute()

cchantep:
I suggest that it would be better to use `SQL("SIMPLE STATEMENT")` if there is no parameter (instead of `BatchSql(...)`.

> Please note that in Anorm master, `BatchSql("STATEMENT", ...)` should be used instead of `BatchSql(SQL("STATEMENT"), ...)`.

optama:
@cchantep I used sql statements in above examples as dummys, in reality, and as an actual use case, one might have a batch insert/update requiring parameters where total count of inserts/updates is dynamic.
Am I understanding correctly in that you suggest one should check for wether the parameter list is empty before using BatchSql?

cchantep:
Batch is done for executing statement `n` times, with `n` being the count of list of list of parameter. If you don't check and `n = 0`, then such batch will just do nothing.

optama:
@cchantep Ok, thank you for the clarification, much appreciated.

cchantep:
@optama you could look at #16 how a first mandatory parameter list is checked (new batch factory).

core/src/test/scala/anorm/H2Database.scala
```diff
@@ -5,11 +5,10 @@ import scala.util.Random
 
 trait H2Database {
 
-  def withH2Database[R](block: Connection => R) = {
-    val dbName = "test" + Random.alphanumeric.take(6).mkString("")
+  def withH2Database[R](block: Connection => R): R = {
+    val url = "jdbc:h2:mem:test" + Random.alphanumeric.take(6).mkString("")
+    val connection = DriverManager.getConnection(url, "sa", "")
 
-    Class.forName("org.h2.Driver")
```
jroper: I would have thought that without this, the driver might not be loaded, and so won't have had a chance to register itself with DriverManager?  Tests are passing, so I guess I'm wrong, but why?

cchantep: Not with Java8. It even blocking.

PR #16, opened 2015-02-12
Do not allow to prepare BatchSql with no list of parameter

cchantep:
Require at least one (first) list of parameters.

```scala
BatchSql("...", Seq[NamedParameter](...), ...)
```


cchantep:
Ready, better after #15

cchantep:
Validated now

cchantep:
Thanks @jroper

PR #17, opened 2015-02-27
Column version for JDBC vendor type wrapping timestamp

cchantep:
Support type like `microsoft.sql.DateTimeOffset`: http://stackoverflow.com/questions/28716932/how-to-convert-microsoft-sql-datetimeoffset-to-date-in-anorm

### DIFF
--- a/core/src/main/scala/anorm/ToStatement.scala
+++ b/core/src/main/scala/anorm/ToStatement.scala
@@ -15,7 +15,7 @@ import java.util.{ UUID => JUUID }
 
 import java.math.{ BigDecimal => JBigDec, BigInteger }
 
-import java.sql.{ PreparedStatement, Types, Timestamp }
+import java.sql.{ PreparedStatement, Timestamp }
 
 /** Sets value as statement parameter. */
 trait ToStatement[A] {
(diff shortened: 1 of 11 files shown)


cchantep:
(Just first draft, doc and similar conversion for Joda and Java7 temporal types should be comited there too)

cchantep:
Validated now

core/src/main/scala/anorm/ToStatement.scala
```diff
@@ -53,8 +53,9 @@ object ToStatement extends JodaToStatement {
    * }}}
    */
   implicit object javaBooleanToStatement extends ToStatement[JBool] {
+    val jdbcType = implicitly[ParameterMetaData[JBool]].jdbcType
     def set(s: PreparedStatement, i: Int, b: JBool): Unit =
-      if (b != null) s.setBoolean(i, b) else s.setNull(i, Types.BOOLEAN)
+      if (b != null) s.setBoolean(i, b) else s.setNull(i, jdbcType)
```
cchantep: Uses meta typeclass to resolve the JDBC type

jroper:
Nice.

cchantep:
Thanks @jroper

PR #18, opened 2015-03-06
SQL parser causes StackOverflowError

cchantep:
The following snippet causes a `StackOverflowError`.

```scala
anorm.SQL(s"select '{${(1 to 1000).mkString(",")}}'::int[]")
```

### DIFF
--- a/core/src/main/scala/anorm/SqlStatementParser.scala
+++ b/core/src/main/scala/anorm/SqlStatementParser.scala
@@ -66,18 +66,14 @@ object SqlStatementParser extends JavaTokenParsers {
 
   private val variable: Parser[TokenGroup] =
     "{" ~> (ident ~ (("." ~> ident)?)) <~ "}" ^^ {
-      case i1 ~ i2 =>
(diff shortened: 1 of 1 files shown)


cchantep:
@carymrobbins Even if a workaround can be applied to avoid such issue, please note that passing parameters in this way is not the recommanded way.

carymrobbins:
@cchantep - True, although, really I am just executing raw SQL and not passing any parameters.  I am doing the interpolation manually since I am dealing with safe values.  Ideally, anorm wouldn't attempt to parse `{..}` as parameters if they reside inside of a string literal.

cchantep:
Anorm will never try to parse parameter from SQL statement, as it is highly dependent on the DB vendor. 

Anorm lets you use native SQL directly, not re-inventing a DB DSL.

If you want to pass parameters along, it's recommanded to use the mecanism provided for (and explained in the doc).

cchantep:
@jroper the attached PR should be fine now

PR #20, opened 2015-03-09
Improve SQL prepration performance

cchantep:
Use local/private mutable `StringBuffer` when generating raw SQL statement.


cchantep:
Thanks @jroper

PR #22, opened 2015-03-28
Row.asList & .asMap are needlessly O(N^2) in the number of columns

cchantep:
[Row.asList](https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/Row.scala#L22) is currently defined as: 
````
lazy val asList: List[Any] = data.foldLeft[List[Any]](Nil) { (l, v) =>
    if (metaData.ms(l.size).nullable) l :+ Option(v) else l :+ v
  }
````
This code loops over `data` (= list with num_columns items in it), then build a transformed result list that is both queried for size (= O(N) each time), and _appended_ to (again O(N) each time). 

We also index into `metaData.ms` (a list with num_columns of items), which is again O(N). 

Since we do O(N) operations O(N) times, this makes the overall time complexity O(N^2).  

Rewriting this to something like: 
````
  lazy val asList: List[Any] = {
    (data, metaData.ms).zipped.map((field, meta) => if (meta.nullable) Option(field) else field)
  }
````
Should resolve the issue and maybe make what's going on a little more clear? 

Similarly for [Row.asMap](https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/Row.scala#L37):
````
  lazy val asMap: Map[String, Any] =
    data.foldLeft[Map[String, Any]](Map.empty) { (m, v) =>
      val d = metaData.ms(m.size)
      val k = d.column.qualified
      if (d.nullable) m + (k -> Option(v)) else m + (k -> v)
    }
````
It indexes `metaData.ms` again, and builds an immutable map one item at a time without a builder / buffer. 

Changing it to: 
````
  lazy val asMap: Map[String, Any] =
    (data, metaData.ms).zipped.map((field, meta) => {
      meta.column.qualified -> (if (meta.nullable) Option(field) else field)
    }).toMap
````
Should resolve the issue. 

(sorry, not good with PRs...)

### DIFF
--- a/core/src/main/scala/anorm/Row.scala
+++ b/core/src/main/scala/anorm/Row.scala
@@ -19,8 +19,8 @@ trait Row {
    *
    * @see #as
    */
-  lazy val asList: List[Any] = data.foldLeft[List[Any]](Nil) { (l, v) =>
-    if (metaData.ms(l.size).nullable) l :+ Option(v) else l :+ v
+  lazy val asList: List[Any] = (data, metaData.ms).zipped.map { (v, m) =>
+    if (m.nullable) Option(v) else v
   }
 
   /**
@@ -34,12 +34,10 @@ trait Row {
    *
    * @see #as
    */
-  lazy val asMap: Map[String, Any] =
-    data.foldLeft[Map[String, Any]](Map.empty) { (m, v) =>
-      val d = metaData.ms(m.size)
-      val k = d.column.qualified
-      if (d.nullable) m + (k -> Option(v)) else m + (k -> v)
-    }
+  lazy val asMap: Map[String, Any] = (data, metaData.ms).zipped.map { (v, m) =>
+    val k = m.column.qualified
+    if (m.nullable) (k -> Option(v)) else k -> v
+  }.toMap
 
   /**
    * Returns row as `T`.
@@ -105,16 +103,9 @@ trait Row {
     }).get // TODO: Safe alternative
 
   // Data per column name
-  private lazy val columnsDictionary: Map[String, Any] = {
-    @annotation.tailrec
-    def loop(meta: List[MetaDataItem], dt: List[Any], r: Map[String, Any]): Map[String, Any] = (meta, dt) match {
-      case (m :: ms, d :: ds) => loop(ms, ds,
-        r + (m.column.qualified.toUpperCase -> d))
-      case _ => r
-    }
-
-    loop(metaData.ms, data, Map.empty)
-  }
+  private lazy val columnsDictionary: Map[String, Any] =
+    (metaData.ms, data).zipped.map((m, v) =>
+      m.column.qualified.toUpperCase -> v).toMap
 
   // Data per column alias
   private lazy val aliasesDictionary: Map[String, Any] = {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jxtps** [comment]
The zipped call is quite light as it just creates a simple wrapper object:
````
    def zipped[El1, Repr1, El2, Repr2]
      (implicit w1: T1 => TraversableLike[El1, Repr1],
                w2: T2 => IterableLike[El2, Repr2]
      ): Tuple2Zipped[El1, Repr1, El2, Repr2] = new Tuple2Zipped((x._1, x._2))
````
`Tuple2Zipped.map` is: 
````
  def map[B, To](f: (El1, El2) => B)(implicit cbf: CBF[Repr1, B, To]): To = {
    val b = cbf(colls._1.repr)
    b.sizeHint(colls._1)
    val elems2 = colls._2.iterator

    for (el1 <- colls._1) {
      if (elems2.hasNext)
        b += f(el1, elems2.next())
      else
        return b.result()
    }

    b.result()
  }
````

Which just creates a new builder and then appends the result of the function, so it's about as efficient as Scala & functional style will permit (and the JIT probably even takes out the creation of the tuple if it's in hot code).  

The foldLeft version creates a bunch of temporary objects (that may also get optimized out, but it's a less clear case) and is also somewhat more complex from the perspective of a human reading the source - the zipped.map style should be about as clear as it is possible to make it...

**Comment by jxtps** [comment]
(Note the important difference between `(a, b).zipped` and `(a zip b)`. The first one is great & light, the second _n'est pas terrible..._)

**Comment by cchantep** [comment]
I feel it's more a question of both style preference and guess about internal optimisation. Is there a benchmark.

**Comment by jxtps** [comment]
Separately it might make sense to add a `lazy val hasNullableColumn = ms.exists(_.nullable)` to `MetaData` and avoid the copy operation (= return `data` directly) in `asList` for the rather common case of no columns actually being nullable (I have many tables in several projects, but not a single nullable column from what I can recall). 

Style vs other considerations: It's not my project, so the project owners will have to make that call. I suspect it could be shown that the `zipped.map` is more performant, but I don't have a specific benchmark for you. Regardless, `zipped.map` feels like "easy" functional programming, whereas `foldLeft` is more "advanced" (certainly takes more lines and more thinking to interpret), and usually simpler code = better, but again, that's not my decision to make here.

**Comment by jxtps** [comment]
Note that for `zipped.map` you don't need the `case` since that map takes a function of two parameters, not a tuple (another difference with `(a zip b).map`).

File: core/src/main/scala/anorm/Row.scala [hunk 1]
```diff
@@ -105,16 +103,9 @@ trait Row {
...
-    }
-
-    loop(metaData.ms, data, Map.empty)
-  }
+  private lazy val columnsDictionary: Map[String, Any] =
+    (metaData.ms, data).zipped.map((m, v) =>
+      m.column.qualified.toUpperCase -> v).toMap
 
   // Data per column alias
   private lazy val aliasesDictionary: Map[String, Any] = {
```
cchantep: According benchmark (with following data), the tailrec impl is more efficient that a `zipped.foldLeft`.

```scala
val meta = (1 to 10000).map(i => MetaDataItem(ColumnName(s"col$i", Some(s"alias$i")), (i % 2 == 0), "java.lang.String")).toList

val data = (1 to 10000).map(i => s"data-$i").toList
```

cchantep:
Should be ok

cchantep:
Could be merged now

cchantep:
Thanks

PR #24, opened 2015-04-11
Provide a shorter syntax to write RowParser for a case class 

cchantep:
I have the following case case and I want to create a RowParser:

```scala
case class History(date: Date, code: String, comment: String)

object History {
  val simple = {
    get[Date]("date") ~ get[String]("code") ~ get[String]("comment") map {
      case date ~ code ~ comment => History(date, code, comment)
    }
  }
}
```

I can use `flatten` to transform `A~B~C` types to `(A,B,C)` and then use `tupled` but the syntax is a bit long and complicated:
```scala
val simple = {
  (get[Date]("date") ~ get[String]("code") ~ get[String]("comment") map flatten) map (History.apply _ tupled)
} 
```
If I don't redefine the companion object it's a little bit better:
```scala
val simple = {
  get[Date]("date") ~ get[String]("code") ~ get[String]("comment") map flatten map History.tupled
}
```

But it would be great if we could reduce the syntax to avoid the double `map`. Do you think this is possible ?

Thanks

### DIFF
--- a/docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md
+++ b/docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md
@@ -614,7 +614,7 @@ val populations: List[String ~ Int] =
   SQL("SELECT * FROM Country").as((str("name") ~ int("population")).*) 
 ```
 
-As you see, this query’s result type is `List[String~Int]` - a list of country name and population items.
+As you see, this query’s result type is `List[String ~ Int]` - a list of country name and population items.
 
 You can also rewrite the same code as:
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala/anorm/Flattener.scala
```diff
@@ -0,0 +1,559 @@
+package anorm
+
+sealed trait Flattener[T] {
```
cchantep: Generalized from former `TupleFlattener` to be able to flatten in any case, whatever is the number of column (including a single one).
Generalisation useful for function application in any case.

docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md
```diff
@@ -639,6 +639,17 @@ val result: List[(String, Int)] =
   SQL("select * from Country").as(parser.flatten.*)
 ```
 
+A `RowParser` can be combined with any function to applied it with extracted columns.
```
Mogztter: Love it ! Thanks :+1:

docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md
```diff
@@ -639,6 +639,17 @@ val result: List[(String, Int)] =
...
+
+```scala
+import anorm.SqlParser.{ int, str, to }
+
+def display(name: String, population: Int): String = 
+  s"The population in $name is of $population."
+
+val parser = str("name") ~ int("population") map (to(display _))
```
cchantep: Note that `display` function must be partially applied with `_` to be used with `to` (cannot write `to(display)`).

cchantep:
Should be ok for review/merge now

cchantep:
Thanks @jroper

PR #29, opened 2015-04-29
Large and stream parameters

cchantep:
(See http://stackoverflow.com/questions/29884763/what-is-the-proper-way-to-insert-update-a-blob-with-play-2-3?noredirect=1#comment48021034_29884763)

### DIFF
--- a/core/src/main/scala/anorm/ToStatement.scala
+++ b/core/src/main/scala/anorm/ToStatement.scala
@@ -26,11 +26,64 @@ trait ToStatement[A] {
   def set(s: PreparedStatement, index: Int, v: A): Unit
 }
 
-/**
- * Provided conversions to set statement parameter.
+/*
+ * Provided instances of `ToStatement` with the lower priority.
+ * Actually it makes sure that `byteArrayToStatement`
+ * (defined in `ToStatementPriority1`) is resolved before `arrayToParameter`.
  */
-object ToStatement extends JodaToStatement {
+sealed trait ToStatementPriority0 {
   import scala.collection.immutable.SortedSet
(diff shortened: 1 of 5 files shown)


cchantep:
Fixed

fosskers:
+1, considering I'm the one who wrote the SO thread.

core/src/main/scala/anorm/ToStatement.scala
```diff
@@ -613,3 +661,25 @@ sealed trait JodaToStatement { // TODO: Move in a separate file
       else s.setNull(index, meta.jdbcType)
   }
 }
+
+sealed trait ToStatementPriority1 extends ToStatementPriority0 {
```
jroper: It would be good to document here why we need different priority implicits, it's really hard to tell just by reading the code.  I'm guessing this is to ensure that `Array[Byte]` takes priority over `Array[T]`.

cchantep: Exactly (also checked by the specs)

cchantep: Updated

cchantep:
Thanks @jroper

fosskers:
What `play` release is this likely to be in?

cchantep:
Next one (2.4). In the meanwhile you can clone this repo to [build it locally](https://github.com/playframework/anorm/blob/master/README.md#build-manually).

fosskers:
Thanks. I'm going to wait for the stable, though. It's client work.

cchantep:
As there is nothing instable in the current master, that more release question.

PR #30, opened 2015-05-07
Switched to using sonatype plugin

jroper:


cchantep:
@play-pull-request-validator

PR #32, opened 2015-05-17
Parser macros

cchantep:
The macros `namedParser[T]` or `indexedParser[T]` (or `parser[T](names)`) can be used to create a `RowParser[T]` at compile-time, for any case class `T`.

```scala
import anorm.{ Macro, RowParser }

case class Info(name: String, year: Option[Int])

val parser: RowParser[Info] = Macro.namedParser[Info]
val result: List[Info] = SQL"SELECT * FROM list".as(parser.*)
```


Migration24.md
```diff
@@ -45,6 +45,19 @@ SQL"""#$cmd * FROM #$table WHERE id = ${"id1"} AND code IN (${Seq(2, 5)})"""
...
+## Macros
+
+The macros `namedParser[T]` or `indexedParser[T]` (or `parser[T](names)`) can be used to create a `RowParser[T]` at compile-time, for any case class `T`.
+
+```scala
+import anorm.{ Macro, RowParser }
+
+case class Info(name: String, year: Option[Int])
+
+val parser: RowParser[Int] = Macro.namedParser[Info]
```
cchantep: Typo `RowParser[Info]`

core/src/main/scala/anorm/Macro.scala
```diff
@@ -0,0 +1,177 @@
...
+  def namedParserImpl_[T: c.WeakTypeTag](c: Context)(names: c.Expr[String]*): c.Expr[Any] = {
+    import c.universe._
+
+    val tpe = c.weakTypeTag[T].tpe
+    val ctor = tpe.declaration(nme.CONSTRUCTOR).asMethod
+    val params = ctor.paramss.flatten
+
+    if (names.size < params.size) {
+      c.abort(c.enclosingPosition,
+        "not column name for parameters: ${show(names)} < $params")
```
cchantep: "no column name ..."

docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md
```diff
@@ -212,6 +212,61 @@ SQL"""#$cmd * FROM #$table WHERE id = ${"id1"} AND code IN (${Seq(2, 5)})"""
...
+## Generated parsers
+
+The macro `namedParser[T]` can be used to create a `RowParser[T]` at compile-time, for any case class `T`.
+
+```scala
+import anorm.{ Macro, RowParser }
+
+case class Info(name: String, year: Option[Int])
+
+val parser: RowParser[Int] = Macro.namedParser[Info]
```
cchantep: `RowParser[Info]`

cchantep:
Should be ok now

cchantep:
@jroper would you mind to have a look at this one?

cchantep:
@jroper I would like this one to be included in 2.4.0. Is it still possible? (seen the new tag added to prepare the release)

jroper:
Hi @cchantep I was thinking we could instead include this in the next version - since it might take a while for the feature to mature, and it also requires compiler plugins if using 2.10.

cchantep:
Next version 2.4.1? The compiler plugins for 2.10 are used to build the macro itself, written with quasiquotes, it doesn't add extra dependency (if it's the concern). Which aspect would you like to be more mature about this one?

jroper:
I think it would just be good to have a release to accompany the Play 2.4.0 release without this.  We can add this to anorm 2.5.0, which we can release in the next few weeks if you want, there's no reason anorm has to match Play release numbers anymore.

cchantep:
Fine for me.

PR #38, opened 2015-05-29
Fix migration documentation

cchantep:
Macro will be release after 2.4


cchantep:
@jroper would you mind to have a look at this one?

PR #39, opened 2015-06-03
Fix interpolated statements with multiple '#$' values

colevk:
Currently, the `SQL` interpolator mangles queries that have multiple values inserted with `#$value`.

    scala> println(SQL"select #${1}, #${2}, #${3}, #${4}, #${5};".getFilledStatement(connection).toString)
    4, 2, select 1, 3, 5

When parsing a token prepended by `#`, the `tokenize` function shouldn't reverse `gts`.

### DIFF
--- a/core/src/test/scala/anorm/StatementParserSpec.scala
+++ b/core/src/test/scala/anorm/StatementParserSpec.scala
@@ -15,11 +15,11 @@ object StatementParserSpec extends org.specs2.mutable.Specification {
   "Statement" should {
     "be parsed with 'name' and 'cat' parameters and support multiple lines" in {
       SqlStatementParser.parse("""
-        SELECT * FROM schema.table 
+        SELECT * FROM schema.table
         -- comment
         WHERE (name = {name} AND category = {cat}) OR id = ?
       """) aka "updated statement and parameters" must beSuccessfulTry(
-        TokenizedStatement(List(TokenGroup(List(StringToken("""SELECT * FROM schema.table 
+        TokenizedStatement(List(TokenGroup(List(StringToken("""SELECT * FROM schema.table
         -- comment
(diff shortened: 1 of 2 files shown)


jroper:
Can you write a test for this?

cchantep:
Thanks for this. As indicated by @jroper, it would be nice to have a test about that.

I would suggest to update the `StatementParserSpec` as following.

```patch
diff --git a/core/src/test/scala/anorm/StatementParserSpec.scala b/core/src/test/scala/anorm/StatementParserSpec.scala
index a8becfc..eb5cb00 100644
--- a/core/src/test/scala/anorm/StatementParserSpec.scala
+++ b/core/src/test/scala/anorm/StatementParserSpec.scala
@@ -113,7 +113,9 @@ object StatementParserSpec extends org.specs2.mutable.Specification {
         case QueryExecution(s, p) => stringList :+ "ko"
       })
 
-      SQL"""#$cmd * FROM #$table WHERE id = ${"id1"} AND code IN (${Seq(2, 5)})""".as(SqlParser.scalar[String].single) must_== "ok"
+      lazy val query = SQL"""#$cmd * FROM #$table WHERE id = ${"id1"} AND code IN (${Seq(2, 5)})"""
+
+      query.sql.stmt aka "tokenized statement" must_== TokenizedStatement(List(TokenGroup(List(StringToken("SELECT"), StringToken(" * FROM "), StringToken("Test"), StringToken(" WHERE id = ")), Some("_0")), TokenGroup(List(StringToken(" AND code IN (")), Some("_1")), TokenGroup(List(StringToken(")")), None)), List("_0", "_1")) and (query.as(SqlParser.scalar[String].single) must_== "ok")
     }
   }
 }
```

colevk:
It looks like the existing test query was parsing correctly entirely by accident, so I added another '#$' term to make sure it failed.

cchantep:
Pls squash the commits as a single one.

Issue #42, opened 2015-06-30
Document changes in member visibility since 2.3


jawshooah:
In a4c9031, the visibility of `Row.metaData` was changed from `public` to `private[anorm]`. However, the 2.4 Migration Guide makes no mention of this change, which was breaking in my case.

Since it was exposed as a public member in previous versions of the API, I imagine I'm not the only one who encountered this issue when upgrading to Play 2.4. It would be helpful to add a note, however brief, that explains the change and how best to refactor around it.

cchantep:
(This change was introduced in Play 2.3.6) What is the case using this value?

jawshooah:
[Are you sure?](https://github.com/playframework/playframework/blob/2.3.6/framework/src/anorm/src/main/scala/anorm/Row.scala) I've been running Play 2.3.7 and haven't had a problem.

The code looks something like this:

```scala
val cells = for (m <- row.metaData.ms) yield {
  val qualifiedName = m.column.qualified
  val rawValue = m.clazz match {
    case "java.lang.Integer" => row[Int](qualifiedName)
    case "java.lang.Long"    => row[Long](qualifiedName)
    ...
  }
  ResultCell(rawValue)
}
```

cchantep:
(See the deprecation annotation in the commit)

It looks like the code is doing `Column` extraction not using `Column`. I guess it should be better to make it as custom instance of `Column` (end of section in doc: https://github.com/playframework/anorm/blob/master/docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md#column-parsers ).

```scala
// having defined `customColum`
val rowParser = get("columnName")(customColumn) ~ ...
```

jawshooah:
Which deprecation annotation? I'm not seeing one for `metaData`.

The problem is that we don't know the names or data types of the columns before retrieving the result set. The application retrieves and transforms arbitrary data from a schema we intentionally know nothing about at run time. I'm not sure how we could create a custom row parser with the `Column` API in this case.

cchantep:
(Deprecation annotation elsewhere in the same commit indicates that's 2.3.6)

You try to parse row without any column contraint ? Column positions or names ? (That's a quite uncommon use case of Anorm for me)

cchantep:
Additional: do you know the number of column to be considered?

jawshooah:
[...]

PR #43, opened 2015-07-01
Add public member metaDataItems to Row trait

jawshooah:
This provides public access to column metadata, which is necessary when the developer does not know the number of columns in the result set or their names and data types, as demonstrated in #42.

### DIFF
--- a/core/src/main/scala/anorm/Row.scala
+++ b/core/src/main/scala/anorm/Row.scala
@@ -9,6 +9,11 @@ trait Row {
   private[anorm] val data: List[Any]
 
   /**
+   * Returns a list of metadata items, one for each column.
+   */
+  lazy val metaDataItems: List[MetaDataItem] = metaData.ms
+
+  /**
    * Returns row as list of column values.
    *
    * {{{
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @jawshooah,

[http://www.typesafe.com/contribute/cla](http://www.typesafe.com/contribute/cla)

core/src/main/scala/anorm/Row.scala
```diff
@@ -9,6 +9,11 @@ trait Row {
   private[anorm] val data: List[Any]
 
   /**
+   * Returns a list of metadata items, one for each column.
+   */
+  lazy val metaDataItems: List[MetaDataItem] = metaData.ms
```
cchantep: I don't think the list of meta should be exposed in this way.

jawshooah: Where is the danger in exposing this information?

In any case, is there a safer way to obtain column metadata? Is it possible to accomplish what I presented in #42 safely?

cchantep:
PR #44 allows to fold over the rows, without exposing the datastructure used internal to manage the metadata.

PR #47, opened 2015-07-09
Short and bytes can be parsed as Int, Long or BigInteger

### LINKED ISSUE #46: Short and Integer Mapping
According to the doc (and the code) a Short can't be parsed as an Int.
As mentioned in the doc that could cause compatibility issues between H2 and Mysql.

I don't understand the reason behind that, especially knowing that an Int can be parsed as a Long.

The workaround is either to use BigDecimal or to define a custom parser.

thibaultdelor:
Fix #46

A column returned as a Short or a Byte can be parsed to an Int, a long or a BigInteger (both java and scala version)

### DIFF
--- a/core/src/test/scala/anorm/ColumnSpec.scala
+++ b/core/src/test/scala/anorm/ColumnSpec.scala
@@ -156,6 +156,18 @@ object ColumnSpec
       SQL("SELECT i").as(scalar[Long].single) aka "parsed long" must_== 4l
     }
 
+    "be parsed from short" in withQueryResult(shortList :+ 3.toShort) {
+      implicit con =>
+        SQL("SELECT s").as(scalar[Long].single).
+          aka("parsed short") must_== 3L
+    }
+
+    "be parsed from byte" in withQueryResult(byteList :+ 4.toByte) {
+      implicit con =>
+        SQL("SELECT b").as(scalar[Long].single).
+          aka("parsed byte") must_== 4L
+    }
+
     "be parsed from false as 0" in withQueryResult(booleanList :+ false) {
       implicit con =>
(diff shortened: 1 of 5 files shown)


typesafehub-validator:
Hi @thibaultdelor,

[http://www.typesafe.com/contribute/cla](http://www.typesafe.com/contribute/cla)

core/src/test/scala/anorm/ColumnSpec.scala
```diff
@@ -156,6 +156,18 @@ object ColumnSpec
       SQL("SELECT i").as(scalar[Long].single) aka "parsed long" must_== 4l
     }
 
+    "be parsed from short" in withQueryResult(shortList :+ 3.toShort) {
+      implicit con =>
+        SQL("SELECT s").as(scalar[Long].single).
+          aka("parsed short") must_== 3l
```
cchantep: Please use `L` as long suffix

thibaultdelor: Sorry I just followed the convention of the file... (it's lower case everywhere else)

cchantep: It shouldn't, as in some editor the `l` can hardly be distinguished from `1`. Thanks.

cchantep:
Thanks @thibaultdelor . Would you mind to also update the [migration guide](https://github.com/playframework/anorm/blob/master/Migration25.md), to indicate there are new conversions for these numeric types (as in the [previous migration guide](https://github.com/playframework/anorm/blob/master/Migration24.md) please). Best regards.

thibaultdelor:
Changed my lower caser l to upper case, updated the migration guide and the highlights

cchantep:
Thanks

PR #49, opened 2015-07-12
Support Joda LocalDate

cchantep:
As parameter and column


core/src/main/scala/anorm/SqlQueryResult.scala
```diff
@@ -42,23 +42,4 @@ final case class SqlQueryResult(
     statement.acquireFor(_.getWarnings).fold[Option[SQLWarning]](
       _.headOption.map(new SQLWarning(_)), Option(_))
 
-  /** Applies current parser with optionnal list of rows (0..n). */
```
cchantep: Cleanup

cchantep:
Thanks @jroper

PR #52, opened 2015-07-22
Wrong exception handling in SqlRequestError as of Anorm 2.4.0?

cchantep:
In Anorm 2.4.0, file anorm/Anorm.scala, lines 19 and 20 it reads:

    /** Returns error as a failure. */
    def toFailure = Failure(sys.error(message))

In Scala 2.11.7, file scala/sys/package.scala, line 27 it reads:

    def error(message: String): Nothing = throw new RuntimeException(message)

Mind the `throw` keyword! Next, in Scala 2.11.7, file scala/Try.scala, line 198 it reads:

    final case class Failure[+T](exception: Throwable) extends Try[T] {

This leads me to the conclusion that a `Failure` object is never returned, but a `RuntimeException` is thrown instead each time execution hits the initial `toFailure` method/function.

I guess this explains why I don't get a reasonable error message sometimes, for example if I forget to provide a parameter to an `SqlQuery`:

    SQL("SELECT content FROM documents WHERE isbn = {isbn}") as myParser.single

Here I forgot to provide the `{isbn}` parameter.

BTW: Wouldn't it be better to use an `SQLException` instead anyway? This would describe the problem domain much better than a generic `RuntimeException`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: scala
 jdk:
   - oraclejdk8
+before_script:
+- mkdir -p $HOME/.sbt/launchers/0.13.8/
+- curl -L -o $HOME/.sbt/launchers/0.13.8/sbt-launch.jar http://dl.bintray.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/0.13.8/sbt-launch.jar
 script:
   - sbt +test docs/test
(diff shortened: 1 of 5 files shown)


cchantep:
Thanks for reporting. You're right there is an issue, fixed by the attached PR. Best regards.

PR #56, opened 2015-08-05
Iteratee module

cchantep:
Get Anorm results as `Enumerator` (see #51); e.g.:

```scala
val Enumerator[String] = Iteratees.from(SQL"SELECT * FROM Test", SqlParser.scalar[String])
```


Groggy:
There is a problem using a `Channel` to feed an `Enumerator` like that : if there is no `Iteratee` consuming the data quickly enough, you will end up with an `OutOfMemoryError` because all the results are loaded in memory.

cchantep:
@Groggy I can't see how your code can consume the result from the returned `Enumerator`:

```scala
val (it, resultEnum): (Iteratee[String, Unit], Enumerator[String]) = Concurrent.joined[String]

it.feed(Input.El("Foo")).flatMap { _ => it.feed(Input.EOF) }

val res: List[String] = Await.result(
  resultEnum.run(Iteratee.getChunks[String]), Duration("5s"))

// res.isEmpty == true
```

Groggy:
@cchantep It's because you don't feed `Input.EOF` to the right `Iteratee`. You need to use the one in the result of the `feed(Input.El("Foo"))`.

```scala
val (it, resultEnum): (Iteratee[String, Unit], Enumerator[String]) = Concurrent.joined[String]

it.feed(Input.El("Foo")).flatMap { _.feed(Input.EOF) }

val res: List[String] = Await.result(
  resultEnum.run(Iteratee.getChunks[String]), Duration("5s"))

// res.isEmpty == List("Foo")
```

cchantep:
@Groggy I don't see memory issue with an unconsumed channel

Groggy:
@cchantep With a `Channel`, there is no feedback to know if data is consumed or not. So the results of the query are pushed to the `Channel` as fast as it can get it from the database, loading everything in memory if nothing consume the data (or nothing fast enough).

Using an `Iteratee`, we can only feed it when it is ready, i.e. the data is consumed, avoiding loading everything in memory if there is no consumer. You can use `Concurrent.buffer` to load more than one result in memory but still avoiding filling the memory if you are afraid that pushing one at a time might slow things.

cchantep:
@Groggy `Channel` is using `Iteratee` and the doc indicates "Iteratees will only receive input sent from the enumerator after they have attached to the broadcasting enumerator", so I do see why it would make it fill memory with unconsumed result. BTW monitoring memory during tests with chunk of 4096, I don't see such issue. Do you have a test case producing such situation?

Groggy:
@cchantep I filled the memory when I did a query returning millions of rows. Within 10 seconds I got an `OutOfMemoryError` and the GC running like hell !

cchantep:
As said, even with big elements fed into the channel, I don't see memory issue.

Groggy:
You're right the issue doesn't appear if no `Iteratee` is attached to the `Enumerator`. But it appears when there is an `Iteratee` consuming data from the `Enumerator` too slowly.

cchantep:
@Groggy You can try:

```scala
    @annotation.tailrec
    def pushToChannel(chan: Concurrent.Channel[T])(cursor: Option[Cursor]): Unit = cursor match {
      case Some(cursor) => cursor.row.as(parser) match {
        case Success(elem) => {
          chan.push(elem)
          pushToChannel(chan)(cursor.next)
        }
        case Failure(err) => chan.end(err)
      }
      case _ => chan.eofAndEnd()
    }

    Concurrent.unicast[T](onStart = { c => sql.withResult(pushToChannel(c)) })
```

Groggy:
Same thing.

You can test with this `Iteratee` :
```scala
def slowIteratee[I]: Iteratee[I, Unit] = {
	def step(input: Input[I]): Iteratee[I, Unit] = input match {
		case Input.El(s) =>
			println(s)

			Iteratee.flatten {
				Future {
					Thread.sleep(1000)
					Cont(step)
				}
			}

		case Input.Empty | Input.EOF =>
			Done((), Input.EOF)
	}
	
	Cont(step)
}
```

cchantep:
As it's not something specific to Anorm, I suggest to ask on the [Play MailingList](https://groups.google.com/forum/#!forum/play-framework).

Groggy:
I agree it's not specific to Anorm, but the goal of this issue is to provide an `Enumerator` of results from a query. The way the `Enumerator` is built by Anorm should not conditioned how it is consumed by the rest of the application.

cchantep:
@Groggy The broadcast or unicast should not introduce the memory issue you mention.

cchantep:
@jroper can you have a look at this one?

build.sbt
```diff
@@ -43,13 +43,25 @@ lazy val anorm = project
...
-      "com.typesafe.play" %% "play-iteratees" % "2.4.0" % Test)
+    ).map("org.specs2" %% _ % "2.4.9" % Test)
   }).dependsOn(`anorm-tokenizer`)
 
+lazy val `anorm-iteratee` = (project in file("iteratee"))
+  .enablePlugins(PlayLibrary)
+  .settings(scalariformSettings: _*)
+  .settings(
+    libraryDependencies ++= Seq(
+      "com.typesafe.play" %% "play-iteratees" % "2.3.9" % "provided",
```
jroper: Shouldn't we use the latest version of Play iteratees here?

cchantep:
@jroper I think it would be nice if this change is compatible down to 2.3.9. As the scope is `provided` it can be used use the latest also.

jroper:
It can use the latest as long as 2.3.9 is binary compatible with 2.4.x, otherwise you're likely to get all sorts of `LinkageErrors`.  I just did a binary compatibility check with the 2.4.x branch against 2.3.9, and here are the breakages:

```
[error]  * object play.api.libs.iteratee.RunQueue#Running does not have a correspondent in new version
[error]  * method apply(java.lang.Object,java.lang.Object)play.api.libs.iteratee.Iteratee in interface play.api.libs.iteratee.Iteratee#EofOrElse's type has changed; was (java.lang.Object,java.lang.Object)play.api.libs.iteratee.Iteratee, is now: (scala.Function0,java.lang.Object)play.api.libs.iteratee.Iteratee
[error]  * abstract method apply(scala.Function0,java.lang.Object)play.api.libs.iteratee.Iteratee in interface play.api.libs.iteratee.Iteratee#EofOrElse does not have a correspondent in old version
[error]  * class play.api.libs.iteratee.RunQueue#Running does not have a correspondent in new version
[error]  * method takeUpTo(Int,scala.Function1)play.api.libs.iteratee.Enumeratee in object play.api.libs.iteratee.Traversable's type has changed; was (Int,scala.Function1)play.api.libs.iteratee.Enumeratee, is now: (Long,scala.Function1)play.api.libs.iteratee.Enumeratee
[error]  * interface play.api.libs.iteratee.RunQueue#State does not have a correspondent in new version
[error]  * synthetic method play$api$libs$iteratee$RunQueue$$deschedule()Unit in class play.api.libs.iteratee.RunQueue does not have a correspondent in new version
```

cchantep:
You're right @jroper , whatever is the iteratee version used, it can lead to issue if the project depending on anorm-iteratee is pulling a more recent & incompatible version of iteratee (e.g. by the transitive way through Play). The document may need to be clearer about that.

cchantep:
Should be ok now

PR #58, opened 2015-08-11
Deprecation cleanup

cchantep:


cchantep:
Should be ok

cchantep:
Thanks @jroper

PR #60, opened 2015-08-19
Referencing the same table twice causes aliased columns to be read incorrectly.

cchantep:
Stackoverflow Question:
http://stackoverflow.com/questions/32062304/mysql-aliased-tables-in-anorm-are-not-recognized

working demo:
https://github.com/narayanjr/anorm_test

If you reference the same table twice in a query, then alias the same columns on those two tables with unique names, the value of the first aliased term (first referenced in mysql) are replaced by the values of the last aliased term.


cchantep:
should be ok now

cchantep:
Thanks @jroper

PR #62, opened 2015-08-31
Support Timestamp in the Long column parser

cchantep:
+ fix pattern matching for `TimestampWrapper1`

### DIFF
--- a/Migration26.md
+++ b/Migration26.md
@@ -39,3 +39,7 @@ import play.api.libs.iteratee._
 def resultAsEnumerator(implicit con: Connection): Enumerator[String] =
   Iteratees.from(SQL"SELECT * FROM Test", SqlParser.scalar[String])
 ```
+
+## Column conversions
+
+- A date/timestamp column can now be read as a `Long`, representing the [epoch](https://en.wikipedia.org/wiki/Unix_time) milliseconds.
--- a/core/src/main/scala/anorm/Column.scala
+++ b/core/src/main/scala/anorm/Column.scala
@@ -6,6 +6,7 @@ package anorm
 import java.io.{ ByteArrayInputStream, InputStream }
 import java.math.{ BigDecimal => JBigDec, BigInteger }
 import java.util.{ Date, UUID }
+import java.sql.Timestamp
 
 import scala.language.reflectiveCalls
 import scala.util.{ Failure, Success => TrySuccess, Try }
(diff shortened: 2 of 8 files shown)


cchantep:
@jroper Should be ok now

Migration26.md
```diff
@@ -38,4 +38,8 @@ import play.api.libs.iteratee._
 
 def resultAsEnumerator(implicit con: Connection): Enumerator[String] =
   Iteratees.from(SQL"SELECT * FROM Test", SqlParser.scalar[String])
-```
\ No newline at end of file
+```
+
+## Column conversions
+
+- A date/timestamp column can now be read as a `Long`.
```
gmethvin: A long representing what?

core/src/main/scala/anorm/package.scala
```diff
@@ -22,6 +22,16 @@ package object anorm {
   /** Structural type for timestamp wrapper. */
   type TimestampWrapper1 = { def getTimestamp: java.sql.Timestamp }
 
+  object TimestampWrapper1 {
+    import scala.language.reflectiveCalls
+
+    def unapply(that: Any): Option[java.sql.Timestamp] = try {
+      Some(that.asInstanceOf[TimestampWrapper1].getTimestamp)
+    } catch {
+      case _: Throwable => None
```
gmethvin: I would much prefer we only catch `NoSuchMethodException` here (which I think is the only thing we'd expect this to throw).

core/src/main/scala/anorm/Column.scala
```diff
@@ -197,7 +197,12 @@ object Column extends JodaColumn with JavaTimeColumn {
       case long: Long => Right(long)
       case s: Short => Right(s.toLong)
       case b: Byte => Right(b.toLong)
-      case bool: Boolean => Right(if (!bool) 0l else 1l)
+      case bool: Boolean => Right(if (!bool) 0L else 1L)
+      case date: Date => Right(date.getTime)
+      case TimestampWrapper1(ts) =>
```
gmethvin: I think something like this would be clearer:
```scala
case TimestampWrapper1(ts) => Right(if (ts == null) null.asInstanceOf[Long] else ts.getTime)
```

Also I think you could write a utility to share the logic among all the code blocks here dealing with dates/timestamps.

cchantep:
Updated

cchantep:
Thanks @jroper

PR #63, opened 2015-09-03
Allow choosing the columns to extract for an insert query

lclement57:
I am using anorm with an Oracle database and need to retrieve the identifier of a row that has been inserted.
Since the value returned by the JDBC driver is the Oracle ROWID when using the `java.sql.Connection.prepareStatement(String sql, int autoGeneratedKeys)` overload, I need to specify the columns that holds the key and whose value I need to retrieve with the `prepareStatement(String sql, String columnNames[])` version of the method

### DIFF
--- a/core/src/main/scala/anorm/SimpleSql.scala
+++ b/core/src/main/scala/anorm/SimpleSql.scala
@@ -3,7 +3,7 @@ package anorm
 import java.sql.Connection
 
 /** Simple/plain SQL. */
-case class SimpleSql[T](sql: SqlQuery, params: Map[String, ParameterValue], defaultParser: RowParser[T], resultSetOnFirstRow: Boolean = false) extends Sql {
+case class SimpleSql[T](sql: SqlQuery, params: Map[String, ParameterValue], defaultParser: RowParser[T], resultSetOnFirstRow: Boolean = false, extractedColumns: Option[Array[String]] = None) extends Sql {
 
   /**
    * Returns query prepared with named parameters.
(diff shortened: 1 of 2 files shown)


cchantep:
Hi,

Why don't you just pass a custom parser to the `.executeInsert(...)` function as [indicated in the documentation](https://playframework.com/documentation/2.4.x/ScalaAnorm):

"""
If you are inserting data that has an auto-generated Long primary key, you can call executeInsert().

val id: Option[Long] = 
  SQL("insert into City(name, country) values ({name}, {country})")
  .on('name -> "Cambridge", 'country -> "New Zealand").executeInsert()

When key generated on insertion is not a single Long, executeInsert can be passed a ResultSetParser to return the correct key.
"""

lclement57:
The result set returned by the query contains only the `oracle.sql.ROWID` for the inserted row, which is [pseudocolumn](http://docs.oracle.com/cd/B19306_01/server.102/b14200/pseudocolumns008.htm) that contains the address of the row.
I could retrieve this value using a `scalar[oracle.sql.ROWID].singleOpt` parser, but then I would have to perform an additional SELECT to retrieve the value of the column that contains the actual primary key.

cchantep:
Hi, @lclement57 I understand, but this is something specific. Such usage can be documented, but I'm not for having it added in the code. Best regards.

cchantep:
@lclement57 after more review I guess it could be added, but with less impact on the core API, by having the column selection as optional parameter for the `.execute{Insert, Update}`, as current generated key parsed default to `Long` parser.

cchantep:
Hi @lclement57 , you can have a look at #80

cchantep:
Fix from #80 is now merged

PR #73, opened 2015-12-05
Macro indexedParser add a way to specify a Prefix

cchantep:
Hello, currently It's really hard to use a Indexed Parser with a join,

currently it would be really useful if there is a way to specify an optional prefix, like when creating the parser we could have something like that:

val parser: RowParser[Demo] = Macro.indexedParser[Demo](4)

Which would result that the index inside the parser adds +4 

i.e.:
```
parserImpl[T](c) { (t, _, i) => q"anorm.SqlParser.get[$t](${i + 1})" }
```

would be:
```
parserImpl[T](c) { (t, _, i) => q"anorm.SqlParser.get[$t](${i + 1 + index})" }
```

### DIFF
--- a/core/src/main/scala/anorm/Macro.scala
+++ b/core/src/main/scala/anorm/Macro.scala
@@ -32,10 +32,18 @@ object Macro {
     }
   }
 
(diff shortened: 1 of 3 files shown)


cchantep:
@jroper Should be ok

schmitch:
Oh man.. I wanted to do that today..

PR #79, opened 2016-01-29
Added a useful Documentation around the offsetParser

schmitch:


cchantep:
Hi, I'm see the point of this new example, and not sure and second snippet is need for `offsetParser`. This may be shared elsewhere. Best regards.

schmitch:
What do you mean exactly?

cchantep:
I mean the example is unclear to me. So for now I would tend not to include it, as there is already an example about.

PR #80, opened 2016-02-07
Select columns from the generated keys on insert

cchantep:
See #63


cchantep:
Should be ok now

cchantep:
Thanks @jroper

PR #83, opened 2016-03-02
Fix NullPointerException when combining parser with `?`

cchantep:


wsargent:
LGTM

cchantep:
Thanks @wsargent

PR #87, opened 2016-03-06
Changed Macro return type for a better experience

schmitch:
Actually the current Macro implementation will return Any, this is aweful when browsing the code inside an IDE since the IDE can't be sure that the Return type is always of type ``RowParser[T]``.

### DIFF
--- a/core/src/main/scala/anorm/Macro.scala
+++ b/core/src/main/scala/anorm/Macro.scala
@@ -2,15 +2,15 @@ package anorm
 
 object Macro {
   import scala.language.experimental.macros
-  import scala.reflect.macros.Context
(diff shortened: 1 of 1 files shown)


cchantep:
Hi. The `c.Expr[Any]` return type inference is related to the whitebox macro behaviour. I'm not fond of adding `asInstanceOf` statement "just" to make IDE happy.

schmitch:
@cchantep actually it's not related to whitebox behavior. I updated it and now it doesn't use a ``asInstanceOf`` statement and still works. It's better since now you don't have a ``Macro.parser[T]`` which will return `Any`. That's too odd.

PR #89, opened 2016-03-07
Updated Dependencies to match the Play 2.5.0 ones

schmitch:
Updated the Dependencies so that they are equal with the ones in Play 2.5, however my question would be rather should we make them 'provided'? (only joda-time and joda-convert) I mean we only touch the outer interfaces so I don't think that this would be a problem at all.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -50,9 +50,9 @@ lazy val anorm = project
   .settings({
     libraryDependencies ++= Seq(
       "com.jsuereth" %% "scala-arm" % "1.4",
-      "joda-time" % "joda-time" % "2.6",
-      "org.joda" % "joda-convert" % "1.7",
-      "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.1",
+      "joda-time" % "joda-time" % "2.9.2",
+      "org.joda" % "joda-convert" % "1.8.1",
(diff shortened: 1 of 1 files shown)


cchantep:
Anorm no longer follows the Play lifecycle. I should be usable without Play, so provided is not an option for me.

schmitch:
Okai actually another thing what about dropping 'joda-time' at some time in the future i.e. make a deprecation timeline.
So that's my PR didn't used 'provided'

cchantep:
Rather than dropping Joda-Time, as it's useful, it could be in a separate module, but for now I have no strong opinion.

schmitch:
@cchantep to make it a seperate module would be great. Currently we are on Java 8 and don't use joda so dropping as much dependencies is great for us, I try to look into it than we could decide if its a good idea.

cchantep:
My point is that a module "just" for Joda types is heavy considering the benefit.

schmitch:
That's true, too. However is there really a need for Joda on JVM 8 based services? I mean actually older stuff still uses Joda and anorm works great without Play! and Java8 however I think that Joda usage should be eliminated as much as possible once you are on Java 8 the question here is just when will anorm drop support for older JVM's.

cchantep:
That's quite opinionated, so better to discuss on the MailingList. Best regards.

PR #91, opened 2016-03-18
namedParser macro with nested case classes

cchantep:
The namedParser macro is eliminating a lot of unnecessary boilerplate. However, its usefulness is somewhat limited when necessary to combine several parsers in nested case classes as e.g. described in the first example here: https://markatta.com/codemonkey/blog/2012/06/16/parsing-results-with-anorm-in-play-framework-2/

For comparison, Json Inception is able to work with nested classes if it finds implicit definitions for Reads/Writes/Formats. Yes, the mechanism is different and there are some challenges that might be overcome, e.g. recursive case class definitions, but shouldn't it be possible?


wsargent:
@cchantep LGTM.  Assuming you want this as a binary compatible patch release per version increment...

cchantep:
Thanks @wsargent

PR #92, opened 2016-04-01
Macros now work recursively for nested case classes

vdichev:
When using the namedMacro and encountering a nested case class, it should also add the fields of this nested case class to a separate parser, and so on on encountering other nested case classes. This even works with a list of columns to fetch, even though it can be a bit awkward for big hierarchies as the list follows the linearized depth-first traversal order.

There are some caveats, of course: I haven't tried this, but I'm pretty certain cyclic case class structures will stack overflow. Also, case classes with the same field names will fetch the same column, but you get what you pay for.

### DIFF
--- a/core/src/main/scala/anorm/Macro.scala
+++ b/core/src/main/scala/anorm/Macro.scala
@@ -2,18 +2,20 @@ package anorm
 
 object Macro {
   import scala.language.experimental.macros
-  import scala.reflect.macros.whitebox
+  import scala.reflect.macros.{ whitebox, Universe }
+  import whitebox.Context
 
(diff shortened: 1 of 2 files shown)


typesafehub-validator:
Hi @vdichev,

[http://www.lightbend.com/contribute/cla](http://www.lightbend.com/contribute/cla)

cchantep:
There is already [a solution](https://github.com/playframework/anorm/pull/91/files).

vdichev:
Right, should have looked, but was a good exercise anyway. I had even forgotten I had submitted that ticket. And using implicit parsers is more flexible than my solution. Thanks!

PR #102, opened 2016-05-16
Fix column resolution for qualified name

cchantep:
In case of a dotted/qualified name (e.g. `table.column`), first try to resolved using the qualified names of the resultset metadata, instead of first trying the alias (default priority for other cases).

### DIFF
--- a/core/src/main/scala/anorm/SqlRequestError.scala
+++ b/core/src/main/scala/anorm/SqlRequestError.scala
@@ -36,7 +36,10 @@ object ColumnNotFound {
     ColumnNotFound(column, row.metaData.availableColumns)
 }
 
-case class UnexpectedNullableFound(message: String) extends SqlRequestError
+case class UnexpectedNullableFound(reason: String) extends SqlRequestError {
+  lazy val message = s"UnexpectedNullableFound($reason)"
+  override lazy val toString = message
+}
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/anorm/SqlRequestError.scala
```diff
@@ -36,7 +36,10 @@ object ColumnNotFound {
     ColumnNotFound(column, row.metaData.availableColumns)
 }
 
-case class UnexpectedNullableFound(message: String) extends SqlRequestError
+case class UnexpectedNullableFound(reason: String) extends SqlRequestError {
```
gmethvin: Was going to backport but I think we probably don't want to change the this API if we did that, right?

cchantep:
The resolution behaviour has never been really documented. Better to doc it for the next release w/o backport I guess.

cchantep:
In fact it's already documented in this way: "If Country.Code is 'First' and CountryLanguage is 'Second', then in previous example code value will be 'Second'. Ambiguity can be resolved using qualified column name, with table name"

gmethvin:
If it's documented to work this way already in 2.5.x, than it's a bugfix, right?

cchantep:
Would say so. It's expected to work in that way since a long time but the case wasn't properly tested.

PR #84, opened 2016-05-21
Naming Scheme for namedParser

cchantep:
It would actually be really great to have a way of defining a NamingSchema to the namedParser.

Usecase: Mostly I encode my PostgreSQL as such:
> customer_id, invoice_mail

While on Scala I use the following schema:
> customerId, invoiceMail


cchantep:
Ok for review

schmitch:
LGTM (tested mostly the functionality)

build.sbt
```diff
@@ -30,6 +30,9 @@ lazy val anorm = project
   .settings(
   scalacOptions += "-Xlog-free-terms",
     binaryIssueFilters ++= Seq(
+      missMeth("anorm.DefaultParameterValue.stringValue"/* deprecated */),
+      missMeth("anorm.ParameterValue.stringValue"/* deprecated */),
+      missMeth("anorm.BatchSql.apply"/* was deprecated */),
```
cchantep: @gmethvin Deprecated in 2.3.8

gmethvin: Okay, but that means this will require a 2.6 (or greater) version of anorm, right? What is anorm's policy about removing methods?

cchantep: For me that me it could have been removed since 2.4.0 as the deprecation was validated by a new major release between.

PR #106, opened 2016-06-17
Runtime Exception when SQL ~ LIKE $param + '%'

cchantep:
### Anorm Version (2.5.x / etc)
2.5.1

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)
windows 10
### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)
jdk 1.8_91 x64
### Library Dependencies
com.microsoft.sqlserver.jdbc.SQLServerDriver
### Expected Behavior
```
val hell = "Hell"
val data2 = DBMS.withConnection( implicit conn =>
      SQL"""
         SELECT * FROM (SELECT 'Hello' AS COL1, 'World' AS COL2) AS MY_TABLE WHERE COL1 LIKE $hell + '%'
      """.as(SqlParser.str(1).single)
    )
```
run without exception
### Actual Behavior

[RuntimeException: No parameter value for placeholder: 2]
image

### Workaround
```
val hell = "Hell" + "%"
val data2 = DBMS.withConnection( implicit conn =>
      SQL"""
         SELECT * FROM (SELECT 'Hello' AS COL1, 'World' AS COL2) AS MY_TABLE WHERE COL1 LIKE $hell
      """.as(SqlParser.str(1).single)
    )
```

PR #100, opened 2016-05-11
Support for oracle.sql.TIMESTAMP with oracle.jdbc.OracleDriver

ljungqvist:

PR #97, opened 2016-04-16
Update specs2

cchantep:

PR #109, opened 2016-07-24
Missing backport of #97, #99, #100, #102 and #106

cchantep:


schmitch:
Actually I can only Squash the 4 commits since there is no way to add them as 4 commits with the current github repo settings, are you fine with that?

cchantep:
I prefer not

schmitch:
then you either create 4 backports or ask that they add the permission to add a "Merge Commit" for this repo.

PR #116, opened 2016-08-15
Support sealed family in macros

cchantep:


schmitch:
Everything looks good, however one question I have, what happens if two implicit row parsers are provided?

cchantep:
There cannot be two implicit row parser. This ambiguity is checked by the compiler anyway.

BTW that's still WIP

cchantep:
Ready for review

core/src/main/scala/anorm/Macro.scala
```diff
@@ -312,6 +596,15 @@ object Macro {
    */
   def offsetParser[T](offset: Int): RowParser[T] = macro offsetParserImpl[T]
 
+  /**
+   * Returns a row parser generated for a sealed class family.
+   * Each direct known subclasses `C` must be provided with an appropriate
+   * `RowParser[C]` in the implicit scope.
```
julienrf: Can you give more details on the way alternatives are discriminated?

cchantep: Documented: https://github.com/playframework/anorm/pull/116/files#diff-eb37b00efed15755d5f4aa5ee58b8474R42

cchantep:
Review welcome

julienrf:
Hey @cchantep, that’s interesting work, however I have the following remarks/questions:

- Can you elaborate on the advantage of writing your own macro implementation over using shapeless? It seems that what your macro does is already available in shapeless. To me, the advantage of using shapeless would be that: the implementation size of `sealedParser` would be reduced by ~80% and I would be able to review the it (I’m not familiar enough with the macros API to review your current implementation). The only drawback I can see is that the error messages might be less precise.
- I believe that you could avoid the introduction of the `DiscriminatorNaming` and `Discriminate` concepts and just reuse a `SqlParser` instead: the `sealedParser` method could take as parameter a `SqlParser` that would return the fully qualified name of the leaf type to parse.

Also, it might be useful to have something that helps writing data consistent with what the parsers expect (something similar to `play.api.libs.json.Format`). Is this something you have in mind?

cchantep:
> Can you elaborate on the advantage of writing your own macro implementation over using shapeless? It seems that what your macro does is already available in shapeless. To me, the advantage of using shapeless would be that: the implementation size of sealedParser would be reduced by ~80% and I would be able to review the it (I’m not familiar enough with the macros API to review your current implementation). The only drawback I can see is that the error messages might be less precise.

Subjectively, I'm sufficiently at ease with standard macros.
It can be experimented in the future, for now for me there is no strong objective opinion to motivate that switch (the existing code is already using standard macros).

> I believe that you could avoid the introduction of the DiscriminatorNaming and Discriminate concepts and just reuse a SqlParser instead: the sealedParser method could take as parameter a SqlParser that would return the fully qualified name of the leaf type to parse.

I guess you mean `RowParser` (`SqlParser` is an utility companion there).
For me parsing row/data is a concern different from such naming one, so having different function kinds makes sense.

> Also, it might be useful to have something that helps writing data consistent with what the parsers expect (something similar to play.api.libs.json.Format). Is this something you have in mind?

Some convenient way to turn instances into `List[NamedParameter]` could be designed, having something similar/coherent with the Anorm interpolation is not as easy to design for me.

PR #110, opened 2016-07-24
Use Akka Stream TestKit in the tests

cchantep:

PR #118, opened 2016-08-16
adds a future ioresult to the akka-stream

### LINKED ISSUE #117: AkkaStream.source should return a Source[T, Future[Done]]
Actually `AkkaStream.source` will return a `Source[T, NotUsed]`.

This is bad since we can't use `.mapMaterializedValue(_.onComplete{ _ => database.close() })` which should cleanup the connection after the stream is done.

schmitch:
## Fixes

Fixes #117

## Purpose

Actually this adds a Promise to the GraphStage so that we can track the success. It also has a counter of all the elements that gone trough it.
Needs rebase if merged after #110

### DIFF
--- a/akka/src/main/scala/anorm/AkkaStream.scala
+++ b/akka/src/main/scala/anorm/AkkaStream.scala
@@ -2,15 +2,19 @@ package anorm
 
 import java.sql.Connection
 
-import scala.concurrent.Future
-
-import akka.NotUsed
-import akka.stream.Materializer
-import akka.stream.scaladsl.Source
+import scala.concurrent.{ Future, Promise }
+import akka.stream.{ MaterializationContext, Materializer, Outlet, SourceShape }
+import akka.stream.impl.SourceModule
+import akka.stream.impl.Stages.DefaultAttributes
+import akka.stream.impl.StreamLayout.Module
+import akka.stream.scaladsl.{ Sink, Source }
+import org.reactivestreams.Publisher
 
 /**
  * Anorm companion for the [[http://doc.akka.io/docs/akka/2.4.4/scala/stream/]].
  *
+ * @define materialization It materializes a [[Future]] of [[Int]] containing the number of rows read from the source upon completion,
+ *                         and a possible exception if row parsing failed.
  * @define sqlParam the SQL query
  * @define materializerParam the stream materializer
  * @define connectionParam the JDBC connection, which must not be closed until the source is materialized.
@@ -19,6 +23,8 @@ import akka.stream.scaladsl.Source
 object AkkaStream {
   /**
    * Returns the rows parsed from the `sql` query as a reactive source.
+   *
+   * $materialization
    *
    * @tparam T the type of the result elements
    * @param sql $sqlParam
@@ -39,10 +45,15 @@ object AkkaStream {
    * def resultSource(implicit m: Materializer, con: Connection): Source[String, NotUsed] = AkkaStream.source(SQL"SELECT * FROM Test", SqlParser.scalar[String], ColumnAliaser.empty)
    * }}}
    */
-  def source[T](sql: => Sql, parser: RowParser[T], as: ColumnAliaser)(implicit m: Materializer, con: Connection): Source[T, NotUsed] = Source.fromGraph(new ResultSource[T](con, sql, as, parser))
+  def source[T](sql: => Sql, parser: RowParser[T], as: ColumnAliaser)(implicit m: Materializer, con: Connection): Source[T, Future[Int]] = {
+    val resultSourceShape = SourceShape[T](Outlet[T]("AnormQueryResult.out"))
+    new Source(new ResultSource[T](con, sql, as, parser, resultSourceShape))
+  }
 
   /**
    * Returns the rows parsed from the `sql` query as a reactive source.
+   *
+   * $materialization
    *
    * @tparam T the type of the result elements
    * @param sql $sqlParam
@@ -50,29 +61,33 @@ object AkkaStream {
    * @param m $materializerParam
    * @param connection $connectionParam
    */
-  def source[T](sql: => Sql, parser: RowParser[T])(implicit m: Materializer, con: Connection): Source[T, NotUsed] = source[T](sql, parser, ColumnAliaser.empty)
+  def source[T](sql: => Sql, parser: RowParser[T])(implicit m: Materializer, con: Connection): Source[T, Future[Int]] = source[T](sql, parser, ColumnAliaser.empty)
 
   /**
    * Returns the result rows from the `sql` query as an enumerator.
    * This is equivalent to `source[Row](sql, RowParser.successful, as)`.
+   *
+   * $materialization
    *
    * @param sql $sqlParam
    * @param as $columnAliaserParam
    * @param m $materializerParam
    * @param connection $connectionParam
    */
-  def source(sql: => Sql, as: ColumnAliaser)(implicit m: Materializer, connnection: Connection): Source[Row, NotUsed] = source(sql, RowParser.successful, as)
+  def source(sql: => Sql, as: ColumnAliaser)(implicit m: Materializer, connnection: Connection): Source[Row, Future[Int]] = source(sql, RowParser.successful, as)
 
   /**
    * Returns the result rows from the `sql` query as an enumerator.
    * This is equivalent to
    * `source[Row](sql, RowParser.successful, ColumnAliaser.empty)`.
    *
+   * $materialization
+   *
    * @param sql $sqlParam
    * @param m $materializerParam
    * @param connection $connectionParam
    */
-  def source(sql: => Sql)(implicit m: Materializer, connnection: Connection): Source[Row, NotUsed] = source(sql, RowParser.successful, ColumnAliaser.empty)
+  def source(sql: => Sql)(implicit m: Materializer, connnection: Connection): Source[Row, Future[Int]] = source(sql, RowParser.successful, ColumnAliaser.empty)
 
   // Internal stages
 
@@ -82,6 +97,31 @@ object AkkaStream {
   import akka.stream.stage.{ GraphStage, GraphStageLogic, OutHandler }
 
   private[anorm] class ResultSource[T](
+      connection: Connection,
+      sql: Sql,
+      as: ColumnAliaser,
+      parser: RowParser[T],
+      shape: SourceShape[T])(implicit mat: Materializer) extends SourceModule[T, Future[Int]](shape) {
+
+    override def create(context: MaterializationContext): (Publisher[T], Future[Int]) = {
+      val result = Promise[Int]()
+      (Source.fromGraph(new AnormResult[T](result, connection, sql, as, parser))
+        .runWith(Sink.asPublisher(false)), result.future)
+    }
+
+    override protected def newInstance(shape: SourceShape[T]): SourceModule[T, Future[Int]] =
+      new ResultSource[T](connection, sql, as, parser, shape)
+
+    override def withAttributes(attr: Attributes): Module =
+      new ResultSource[T](connection, sql, as, parser, amendShape(attr))
+
+    override def attributes: Attributes = Attributes.name("resultSource") and DefaultAttributes.IODispatcher
+
+    override protected def label: String = s"ResultSource($parser)"
+  }
+
+  private[anorm] class AnormResult[T](
+      result: Promise[Int],
       connection: Connection,
       sql: Sql,
       as: ColumnAliaser,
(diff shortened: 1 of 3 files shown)


cchantep:
Hi, why not using `Sink.onComplete` (which doesn't require any specific on the `Source` side)?

akka/src/test/scala/anorm/AkkaStreamSpec.scala
```diff
@@ -26,8 +21,8 @@ class AkkaStreamSpec extends org.specs2.mutable.Specification {
     "expose the query result as source" in { implicit ee: EE =>
       withQueryResult(stringList :+ "A" :+ "B" :+ "C") { implicit con =>
         AkkaStream.source(SQL"SELECT * FROM Test", SqlParser.scalar[String]).
-          runWith(Sink.seq[String]) must beEqualTo(Seq("A", "B", "C")).
-          await(0, 5.second)
+            runWith(Sink.seq[String]) must beEqualTo(Seq("A", "B", "C")).
```
cchantep: weird formatting difference, could you check `sbt scalariformFormat test:scalariformFormat`

schmitch:
Since some framworks needing a Source (play) where I can't control the sink ;)

akka/src/test/scala/anorm/AkkaStreamSpec.scala
schmitch: yeah there is also a pretty bad mistake

cchantep:
I don't see the case, neither how the proposed changes would fix such case.

In any case you should be able to use `alsoTo`?

akka/src/main/scala/anorm/AkkaStream.scala
```diff
@@ -72,19 +78,46 @@ object AkkaStream {
...
+  import akka.stream.{ Attributes, Outlet, SourceShape }
+
+  import scala.util.{ Failure, Success }
 
   private[anorm] class ResultSource[T](
       connection: Connection,
       sql: Sql,
       as: ColumnAliaser,
+      parser: RowParser[T],
+      shape: SourceShape[T])(implicit mat: Materializer) extends SourceModule[T, Future[IOResult]](shape) {
```
cchantep: For now I don't see benefit of a new/case specific handling.

Handling the termination, possibly counting the processed element can already be done.

```scala
scala> source.via(Flow.apply.fold(0) { (c, _) => c + 1 }) runWith Sink.last
res1: scala.concurrent.Future[Int]
```

schmitch:
With `FileIO` you can do:
```
Ok.chunked(FileIO.fromPath(path).mapMaterializedValue(_.onComplete{ _ =>
// do something when file is read
})
```

With this patch you can directly chunk data from the database to the Request and close the database after every row is read. This helps when working on big CSV data.
Currently this patch is wrong since I added it to onDownStreamFinish, it should actually be before complete(out)

akka/src/main/scala/anorm/AkkaStream.scala
schmitch: the counting was just mimic from `FileIO`.

akka/src/main/scala/anorm/IOResult.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import scala.util.{ Failure, Try }
+
+/**
+ * Holds a result of an IO operation.
+ *
+ * @param count Numeric value depending on context, for example IO operations performed or bytes processed.
+ * @param status Status of the result. Can be either [[akka.Done]] or an exception.
+ */
+final case class IOResult private[anorm] (count: Long, status: Try[Done]) {
```
cchantep: Any way, the Akka [`IOResult`](http://doc.akka.io/api/akka/2.4.8/?_ga=1.256176713.331620270.1455976396#akka.stream.IOResult) can be used

akka/src/main/scala/anorm/Done.scala
```diff
@@ -0,0 +1,18 @@
...
+package anorm
+
+/**
+ * Typically used together with `Future` to signal completion
+ * but there is no actual value completed. More clearly signals intent
+ * than `Unit` and is available both from Scala and Java (which `Unit` is not).
+ */
+sealed abstract class Done
+
+case object Done extends Done {
```
cchantep: Akka [`Done`](http://doc.akka.io/api/akka/2.4.8/?_ga=1.256176713.331620270.1455976396#akka.Done)  is available

akka/src/main/scala/anorm/IOResult.scala
schmitch: I wanted to use that, but it's `private[akka]` -1

akka/src/main/scala/anorm/AkkaStream.scala
```diff
@@ -2,18 +2,21 @@ package anorm
...
-import akka.stream.scaladsl.Source
+import scala.concurrent.{ Future, Promise }
 
 /**
  * Anorm companion for the [[http://doc.akka.io/docs/akka/2.4.4/scala/stream/]].
  *
- * @define sqlParam the SQL query
- * @define materializerParam the stream materializer
- * @define connectionParam the JDBC connection, which must not be closed until the source is materialized.
+ * @define sqlParam           the SQL query
```
cchantep: weird formatting

akka/src/main/scala/anorm/AkkaStream.scala
```diff
@@ -72,19 +78,46 @@ object AkkaStream {
...
 
   private[anorm] class ResultSource[T](
       connection: Connection,
       sql: Sql,
       as: ColumnAliaser,
+      parser: RowParser[T],
+      shape: SourceShape[T])(implicit mat: Materializer) extends SourceModule[T, Future[IOResult]](shape) {
+
+    override def create(context: MaterializationContext): (Publisher[T], Future[IOResult]) = {
+      val ioResultPromise = Promise[IOResult]()
```
cchantep: As for naming, I don't think the `Promise` suffix is required about that

akka/src/main/scala/anorm/AkkaStream.scala
```diff
@@ -72,19 +78,46 @@ object AkkaStream {
...
+    override def withAttributes(attr: Attributes): Module =
+      new ResultSource[T](connection, sql, as, parser, amendShape(attr))
+
+    override def attributes: Attributes = Attributes.name("resultSource") and DefaultAttributes.IODispatcher
+
+    override protected def label: String = s"ResultSource($parser)"
+  }
+
+  private[anorm] class ResultSourceGraphStage[T](
+      ioResultPromise: Promise[IOResult],
```
cchantep: `result: Promise[IOResult]`

akka/src/main/scala/anorm/AkkaStream.scala
```diff
@@ -72,19 +78,46 @@ object AkkaStream {
...
+
+    override def withAttributes(attr: Attributes): Module =
+      new ResultSource[T](connection, sql, as, parser, amendShape(attr))
+
+    override def attributes: Attributes = Attributes.name("resultSource") and DefaultAttributes.IODispatcher
+
+    override protected def label: String = s"ResultSource($parser)"
+  }
+
+  private[anorm] class ResultSourceGraphStage[T](
```
cchantep: `class AnormResult`

akka/src/test/scala/anorm/AkkaStreamSpec.scala
```diff
@@ -31,6 +26,18 @@ class AkkaStreamSpec extends org.specs2.mutable.Specification {
       }
     }
 
+    "be done if the stream run through" in { implicit ee: EE =>
+      @volatile var applied = false
+      withQueryResult(stringList :+ "A" :+ "B" :+ "C") { implicit con =>
+        Await.result(AkkaStream.source(SQL"SELECT * FROM Test", SqlParser.scalar[String]).
```
cchantep:
```scala
AkkaStream.source(SQL"SELECT * FROM Test", SqlParser.scalar[String]).
  runWith(Sink.last).flatMap(identity) must beLike[IOResult] {
    case io => ???
  }.await(0, 3.seconds)
```

akka/src/main/scala/anorm/IOResult.scala
cchantep: So just using `Future[Int]` as materialization could be sufficient (for now either it's successful with a count, or it's a `Failure`). That would avoid such code duplication.

akka/src/test/scala/anorm/AkkaStreamSpec.scala
schmitch: actually that doesn't work since I want the materialized value and not the return value of the sink.

schmitch:
I applied all your suggestions except the test, which I slightly modified to get the materialized value.

akka/src/test/scala/anorm/AkkaStreamSpec.scala
```diff
@@ -31,6 +26,15 @@ class AkkaStreamSpec extends org.specs2.mutable.Specification {
       }
     }
 
+    "be done if the stream run through" in { implicit ee: EE =>
+      withQueryResult(stringList :+ "A" :+ "B" :+ "C") { implicit con =>
+        AkkaStream.source(SQL"SELECT * FROM Test", SqlParser.scalar[String]).
+          toMat(Sink.ignore)(Keep.left).run() must beLike[Int] {
```
cchantep: `toMat(Sink.ignore)(Keep.left).run() must beEqualTo(3).await(0, 3.seconds)`

akka/src/main/scala/anorm/AkkaStream.scala
```diff
@@ -50,7 +56,7 @@ object AkkaStream {
    * @param m $materializerParam
    * @param connection $connectionParam
```
cchantep: Would be nice to update the Scaladoc to indicate the meaning of the materialized value

schmitch: wouldn't be better to add that to the docs? Like:
```
If you need to close the connection after all rows are materialized you can just map the materialized value:
source.mapMaterializedValue(_.onComplete { _ => connection.close() }) 
```

cchantep: Yes, and would be nice to use a Scaladoc `@define` to make sure this same explanation is used for all the functions

cchantep:
Please squash the commits

akka/src/main/scala/anorm/AkkaStream.scala
```diff
@@ -125,13 +166,18 @@ object AkkaStream {
...
             }
-            case Failure(cause) => fail(out, cause)
+            case Failure(cause) =>
+              result.failure(cause)
+              fail(out, cause)
           }
 
-          case _ => complete(out)
+          case _ =>
+            result.trySuccess(counter)
```
cchantep: I don't think `trySuccess` is needed instead of `success`, as it's a pure value

docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md
```diff
@@ -453,19 +453,38 @@ import java.sql.Connection
...
 
-def resultSource(implicit m: Materializer, con: Connection): Source[String, NotUsed] = AkkaStream.source(SQL"SELECT * FROM Test", SqlParser.scalar[String], ColumnAliaser.empty)
+def resultSource(implicit m: Materializer, con: Connection): Source[String, Future[Int]] = AkkaStream.source(SQL"SELECT * FROM Test", SqlParser.scalar[String], ColumnAliaser.empty)
 
 def countStrings()(implicit m: Materializer, con: Connection): Future[Int] =
   resultSource.runWith(
     Sink.fold[Int, String](0) { (count, str) => count + str.length })
 ```
 
+It materializes a [[Future]] of [[Int]] containing the number of rows read from the source upon completion,and a possible exception if row parsing failed.
```
cchantep:
```
It materializes a `Future` containing either the number of read rows from the source if successful, or the exception if row parsing failed.
```

docs/manual/working/scalaGuide/main/sql/ScalaAnorm.md
```diff
@@ -453,19 +453,38 @@ import java.sql.Connection
...
+import scala.concurrent.Future
+
+import akka.stream.Materializer
+import akka.stream.scaladsl.{ Sink, Source }
+
+import anorm._
+
+def source(implicit m: Materializer, connection: Connection): Source[String, Future[Int]]#ReprMat[String, Unit] = 
+  AkkaStream.source(SQL"SELECT * FROM Test", SqlParser.scalar[String], ColumnAliaser.empty)
+    .mapMaterializedValue(_.onComplete{ _ =>
```
cchantep: `onComplete { _ =>`

cchantep:
Thanks @schmitch

Issue #122, opened 2016-09-08
Lower-case aliases pick same column in joins with multiple references to same table


vdichev:
Anorm Version (2.5.x)

### Expected Behavior

In a query where the same table is joined with itself, aliases don't work properly.

1. Execute e.g. query `select t1.name as name1, t2.name as name2 from test t1 join test t2 on t1.id != t2.id`
2. Refer to a Row returned by query.
3. Observe that `row.apply[String]("name1")` is always equal to `row.apply[String]("name2")`.

### Actual Behavior

1. Same as above
2. Same as above
3. Observe that `row.apply[String]("name1")` is different from `row.apply[String]("name2")`.

Note that it works as expected when aliases in query are upper-case:

```sql
select t1.name as NAME1, t2.name as NAME2 from test t1 join test t2 on t1.id != t2.id
```

### Analysis and proposed solution

The problem can be traced to the `get` method in Row.scala:

```scala
    data <- MayErr(m.column.alias.flatMap(aliasesDictionary.get(_)).
      orElse(columnsDictionary.get(m.column.qualified.toUpperCase())).
```

So the qualified name is upper-cased before being retrieved from the `columnsDictionary`, but the alias is not upper-cased when retrieved from the `aliasesDictionary`. This would also explain how it works when we have the table referenced only once- it falls back to the qualified name. However when there are multiple references to a column in a table, it always picks the first one it encounters, bypassing aliases. I think the alias then should either be upper-cased upon constructing the `MetaDataItem` or, to be more consistent with how `qualified` is used in the code fragment, to be changed to this:

```scala
    data <- MayErr(m.column.alias.flatMap(a => aliasesDictionary.get(a.toUpperCase())).
```

vdichev:
Let me know if you need a PR

cchantep:
Hi, PR is welcome. Please make sure to add appropriate tests so it can be validated. Best regards.

vdichev:
https://github.com/playframework/anorm/pull/125

PanzerKunst:
Thank you for raising this issue, and creating a pull request. I am experiencing a related issue: a bug when selecting twice the same column of a given table (for example when joining twice on that table).

Sample code to illustrate:

    val query = """select d.id as order_id,
      u.id as customer_id, u.prenume as customer_first_name,
      r.prenume as rater_first_name
    from documents d
      inner join useri u on u.id = d.added_by
      inner join useri r on r.id = d.assign_to;"""

[...]

PR #125, opened 2016-09-14
Fix lower-case aliases when table referred multiple times

vdichev:
Fixes https://github.com/playframework/anorm/issues/122

### DIFF
--- a/core/src/test/scala/anorm/H2Database.scala
+++ b/core/src/test/scala/anorm/H2Database.scala
@@ -5,7 +5,7 @@ import scala.util.Random
 
 trait H2Database {
   def withH2Database[R](block: Connection => R): R = {
-    val url = "jdbc:h2:mem:test" + Random.alphanumeric.take(6).mkString("")
+    val url = "jdbc:h2:mem:test" + Random.alphanumeric.take(6).mkString("") + ";DATABASE_TO_UPPER=false"
     val connection = DriverManager.getConnection(url, "sa", "")
 
     try {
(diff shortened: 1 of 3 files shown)


core/src/test/scala/anorm/H2Database.scala
```diff
@@ -5,7 +5,7 @@ import scala.util.Random
 
 trait H2Database {
   def withH2Database[R](block: Connection => R): R = {
-    val url = "jdbc:h2:mem:test" + Random.alphanumeric.take(6).mkString("")
+    val url = "jdbc:h2:mem:test" + Random.alphanumeric.take(6).mkString("") + ";DATABASE_TO_UPPER=false"
```
cchantep: I prefer not to change that, to run the integration tests against the default options, and to provide the unit test for this PR using `withLabel` as suggested

vdichev: After some research into acolyte, it doesn't seem like one can specify a different qualified name and alias using `withLabel`. We need to have a setup with 2 columns with the same qualified name, but *different aliases* in order to get a failing test. Otherwise the test will succeed even without the suggested change and doesn't prove anything.

When the alias is not found, it will fall back to using the qualified name, which will work fine *provided* there is a *single* column with this qualified name.

Please advise how to proceed.

cchantep: The tests are failing for now

cchantep:
Issue #122 is now closed

PR #130, opened 2016-10-13
Correctly parse aliased columns with duplicate qualified names

### LINKED ISSUE #122: Lower-case aliases pick same column in joins with multiple references to same table

mhzajac:
Fixes #122

This makes the same change as #125 , but it didn't seem like constructing queries was really necessary here. The problem was with `Row#get`, so I just constructed some mock rows and meta data.

### DIFF
--- a/core/src/main/scala/anorm/Row.scala
+++ b/core/src/main/scala/anorm/Row.scala
@@ -132,7 +132,7 @@ trait Row {
           orElse(m.column.alias.flatMap(aliasesDictionary.get(_)))
 
       } else {
-        m.column.alias.flatMap(aliasesDictionary.get(_)).
+        m.column.alias.flatMap(a => aliasesDictionary.get(a.toUpperCase)).
           orElse(columnsDictionary.get(m.column.qualified.toUpperCase))
       }
 
--- a/core/src/test/scala/anorm/RowSpec.scala
+++ b/core/src/test/scala/anorm/RowSpec.scala
@@ -37,6 +37,24 @@ class RowSpec extends org.specs2.mutable.Specification {
           aka("column list") must_== List(Some("str"))
 
       }
+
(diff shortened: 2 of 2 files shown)


mhzajac:
Updated.

PR #133, opened 2016-11-03
Add support for Scala 2.12.0

cchantep:


project/plugins.sbt
```diff
@@ -1,11 +1,12 @@
 resolvers ++= DefaultOptions.resolvers(snapshot = true)
 
 addSbtPlugin("com.typesafe.play" % "interplay" % "1.1.2")
-addSbtPlugin("com.typesafe.play" % "play-docs-sbt-plugin" % sys.props.getOrElse("play.version", "2.5.0"))
+addSbtPlugin("com.typesafe.play" % "play-docs-sbt-plugin" % sys.props.getOrElse("play.version", "2.5.2"))
```
marcospereira: Why not Play 2.5.8 as declared above?

build.sbt
```diff
@@ -5,9 +5,9 @@ import com.typesafe.tools.mima.plugin.MimaKeys.{
   binaryIssueFilters, previousArtifacts
 }
 
-val PlayVersion = playVersion(sys.props.getOrElse("play.version", "2.5.4"))
+val PlayVersion = playVersion(sys.props.getOrElse("play.version", "2.5.8"))
```
schmitch: well here it should be `play.version` 2.5.9 as well, I guess, however I don't any place where you even use PlayVersion anymore.

cchantep: Updated

PR #137, opened 2016-11-26
Show pretty string representation about query

cchantep:
See [usage](https://github.com/playframework/anorm/compare/master...cchantep:feature/show?expand=1#diff-699333be0e769d0742891e5d5502a6c3R524)


cchantep:
Thx

PR #145, opened 2017-02-16
Upgrade to Play 2.6.0-M1

marcospereira:
## Purpose

Upgrade play to version 2.6.0-M1.

### DIFF
--- a/docs/build.sbt
+++ b/docs/build.sbt
@@ -2,7 +2,8 @@ name := "anorm-docs"
 
 libraryDependencies ++= Seq(
   component("play-jdbc") % "test",
-  component("play-specs2") % "test"
+  component("play-specs2") % "test",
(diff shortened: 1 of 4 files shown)


docs/build.sbt
```diff
@@ -2,7 +2,8 @@ name := "anorm-docs"
 
 libraryDependencies ++= Seq(
   component("play-jdbc") % "test",
-  component("play-specs2") % "test"
+  component("play-specs2") % "test",
+  "com.h2database" % "h2" % "1.4.193"
```
cchantep: Why is it now required?

marcospereira: H2 used to be a dependency of Play and then it was added transitively here. But it was removed from Play and now should be added explicitly. Keep in mind this is the docs sub project.

PR #147, opened 2017-02-24
PostgreSQL module

cchantep:
Optional/extra module dedicated to PostgreSQL support:

- JSONB support, using Play JSON
- UUID support

## Integration tests (using Docker)

Setup a test DB named 'anorm147' from CLI:

    docker run --rm --name pganorm147 -p ':5432' -e 'POSTGRES_USER=anorm147' -e 'POSTGRES_PASSWORD=anorm147' -e 'POSTGRES_DB=anorm147' postgres:9

In a separate shell, note the port to access the containerized DB:

    docker port pganorm147 | cut -d ':' -f 2

Then create a test table (with the 'anorm147' password):

```
export MY_PG_PORT="32772"
cat > /dev/stdout << EOF | /path/to/bin/psql -h localhost -p "$MY_PG_PORT" -U anorm147 anorm147
CREATE TABLE test(id VARCHAR PRIMARY KEY, json JSONB);
CREATE TABLE test_seq(id UUID PRIMARY KEY);
EOF
```

From the Anorm repository with the branch checkout'ed, execute `sbt`:

```
sbt> project anorm-postgres
sbt> console
```

Edit the following test code (the port) and paste it in the started Scala console.

```scala
import java.util.UUID
import java.sql._
import anorm._, postgresql._
import play.api.libs.json._

val port = 32772
val url = s"jdbc:postgresql://localhost:$port/anorm147"
implicit val con = DriverManager.getConnection(url, "anorm147", "anorm147")

sealed trait MyEnum
case object Bar extends MyEnum
case object Lorem extends MyEnum

// 1. Insert JSON
implicit val w: Writes[MyEnum] = Writes[MyEnum] {
  case Bar => Json.obj("bar" -> 1)
  case Lorem => Json.obj("lorem" -> 2)
}

// 1.a JsValue
SQL"""INSERT INTO test(id, json) VALUES (${"foo"}, ${Json.obj("bar" -> 1)})""".executeUpdate()

// 1.b T: Writes
SQL"INSERT INTO test(id, json) VALUES(${"bar"}, ${asJson[MyEnum](Lorem)})".executeUpdate()

// 2. Insert UUID
SQL"INSERT INTO test_seq VALUES(${UUID.randomUUID()})".executeUpdate()

// 3. Select JSON
implicit val r: Reads[MyEnum] = Reads[MyEnum] { js =>
  (js \ "bar").validate[Int].map {
    case 1 => Bar
    case _ => Lorem
  }
}

// 3.a JsValue
SQL"""SELECT json FROM test WHERE id = ${"foo"}""".
  as(SqlParser.scalar[JsValue].single)

// 3.b JsObject
SQL"""SELECT json FROM test WHERE id = ${"foo"}""".
  as(SqlParser.scalar[JsObject].single)

// 3.c T: Reads
SQL"""SELECT json FROM test WHERE id = ${"foo"}""".
  as(SqlParser.scalar(fromJson[MyEnum]).single)

// 4. Select UUID
SQL"SELECT * FROM test_seq".as(SqlParser.scalar[UUID].*)
```


cchantep:
@gmethvin would the new module require anything special for the next release?

build.sbt
```diff
@@ -68,14 +76,18 @@ lazy val anorm = project
       // New functions
       missMeth("anorm.Sql.unsafeStatement$default$2"),
       missMeth("anorm.Sql.unsafeResultSet"),
-      missMeth("anorm.Sql.unsafeStatement")),
+      missMeth("anorm.Sql.unsafeStatement"),
+      missMeth("anorm.ToStatement.contramap"),
```
gmethvin: Given the fact that `ToStatement` is [documented](https://www.playframework.com/documentation/2.5.x/ScalaAnorm#Custom-parameter-conversions) as a feature, it's very likely this is being used in other libraries, so binary compatibility issues are likely. The same is true for `Column`.

We could work around this by defining a separate implicit class for the ops:
```scala
implicit class ToStatementOps[A](self: ToStatement[A]) {
  final def contramap[B](f: B => A): ToStatement[B] = ToStatement[B] { (s, i, b) => self.set(s, i, f(b)) }
}
```

core/src/main/scala/anorm/Column.scala
```diff
@@ -12,16 +12,87 @@ import scala.util.{ Failure, Success => TrySuccess, Try }
...
+   * def find(id: String) =
+   *   SQL"SELECT enum_code FROM my_table WHERE id = $id".
+   *     as(SqlParser.scalar(myEnumCol).single)
+   * }}}
+   */
+  def map[B](f: A => B): Column[B] = mapResult[B] { a =>
+    try {
+      Right(f(a))
+    } catch {
+      case cause: Throwable => Left(SqlRequestError(cause))
```
gmethvin: All of these catch blocks should only catch `NonFatal(cause)`. If it's a fatal error, I would expect the error to bubble up. I believe this is what most Scala users expect since this is what `Try` does.

build.sbt
```diff
@@ -106,6 +119,18 @@ lazy val `anorm-akka` = (project in file("akka"))
       "com.typesafe.akka" %% "akka-stream-contrib" % "0.6" % Test)
   )).dependsOn(anorm)
 
+lazy val pgVer = sys.props.get("postgres.version").getOrElse("9.4.1212")
+lazy val jsonVer = sys.props.get("playJson.version").getOrElse("2.6.0-M3")
```
gmethvin: Do you need any of the new features of play-json? This might cause binary compatibility issues if you're using anorm with Play 2.5.x.

build.sbt
cchantep: These are new functions, so there is no backward compatibility issue.

On the other side I'm not for using an extra implicit conversion for "magic".

build.sbt
cchantep: Only `Reads.reads` and `Writes.writes`, and only for the new separate module.

core/src/main/scala/anorm/Column.scala
cchantep: Ok

build.sbt
gmethvin: In Scala 2.11, binary compatibility for traits is the same as for interfaces in Java. The trait is compiled to a `ToStatement` interface and a `ToStatement$class` static class with method implementations. When a class extends the trait, the compiler generates the "stub" implementations to forward to the static class. If you add a method to the trait with an implementation, all classes extending that trait need to be recompiled to get the implementation.

So unfortunately it's not binary compatible, which is why I suggested an alternative option.

cchantep: I don't think that the case of a project changing the Anorm JARs without recompiling should be supported.

Anorm (and Play modules in general I think) never claims to have backward compatibility on the ABI.

Whereas it should for the API, which is ok for new functions.

gmethvin: In general with Play, our policy is to avoid breaking the ABI in minor releases, except to fix serious bugs. Slick has a similar policy. It's extremely common for external modules to not upgrade their Play version between releases. So this either causes breakage, or prevents people from upgrading until they can get all modules recompiled for the new version.

Hard to say whether anorm will have the same problems, but if it has any significant number of users it likely will.

Often, we'll have two different implementations of the same change, one that breaks binary compatibility in the future release branch (currently 2.6.x) and another that preserves binary compatibility on the stable release branch (2.5.x). To me this is a good compromise: we get to eventually have the code the way we want it, but preserve binary compatibility for existing users.

cchantep: To trigger such issue, a project would have to replace the Anorm JAR without recompiling, which again for me cannot be anything but a bad practice.

gmethvin: That's not true. Most commonly the conflict occurs through transitive dependencies. e.g. your project uses version 2.5.4 but you have an external lib using 2.5.3. SBT will by default pick the newer version, making the external library unusable. Actually, there are already a number of libraries that haven't updated their anorm version to latest:
 - https://github.com/mhzajac/anorm-relational
 - https://github.com/lucidsoftware/relate
 - https://github.com/giabao/1pay

Sure, in a perfect world, everyone would update and release their library as soon as a new release of any dependency comes out, but that can be quite time consuming to do, and it's just not reasonable to expect that all the time. That's why we have binary compatibility guarantees.

cchantep: > everyone would update and release their library as soon as a new release of any dependency comes out

Or rather use `provided` scope to allow to avoid dependency hell.

Even if I would not guarantee this edge case, I can see the point, and this case better to bump to 2.6.0 finally.

The same issue with transitive deps could still happen then, but if we agree the guarantee does not apply in this case.

gmethvin: Yes, I think it's ok to break binary compatibility in 2.6.0.

`provided` doesn't solve the problem, since the old anorm version would still need to be used for compilation.

cchantep: If Ver1 provides function F1, Ver2 provides same function F1 plus a new function F2, for a project using a tier library depending only on F1, the compilation will succeed with either Ver1 or Ver2.

gmethvin: Not sure I understand your point. If you recompile the library of course it will work. I'm not saying anything won't compile. The problem, like I said, is if you use a library like anorm-relational that's compiled with an older anorm version, its implementations of the `ToStatement` trait would be incomplete. This would likely cause an error at runtime.

cchantep: Only if the tier library force the Anorm version, which won't be the case if properly defined as provided

cchantep:
> The Play JSON dependency is potentially problematic, though, since this also restricts the versions of Play that are binary compatible, so maybe it makes sense to split that out as a separate library.

This is already a separate module

core/src/main/scala/anorm/Column.scala
cchantep: Updated

cchantep:
Updated

cchantep:
Up

build.sbt
```diff
@@ -106,6 +119,23 @@ lazy val `anorm-akka` = (project in file("akka"))
       "com.typesafe.akka" %% "akka-stream-contrib" % "0.6" % Test)
   )).dependsOn(anorm)
 
+lazy val pgVer = sys.env.get("POSTGRES_VERSION").getOrElse("9.4.1212")
+
+val jsonVer = Def.setting[String] {
```
gmethvin: Can we just always depend on 2.6.x? I thought we were releasing 2.6.0 so we can break binary compatibility.

core/src/main/scala/anorm/Column.scala
gmethvin: Am I still looking at an older version of this PR? I thought you updated this.

build.sbt
cchantep: There is no need

core/src/main/scala/anorm/MayErr.scala
```diff
@@ -1,36 +0,0 @@
-/*
- * Copyright (C) 2009-2013 Typesafe Inc. <http://www.typesafe.com>
- */
-package anorm
-
-// TODO: Remove
-
-private[anorm] case class MayErr[+E, +A](toEither: Either[E, A]) {
```
richdougherty: Did you just remove this to simplify the implementation by using `Either` instead?

postgres/src/main/scala/package.scala
```diff
@@ -0,0 +1,127 @@
...
+  implicit val jsValueColumn: Column[JsValue] =
+    Column.nonNull[JsValue] { (value, meta) =>
+      @inline def str: Option[String] = value match {
+        case o: PGobject => Some(o.getValue)
+        case s: String => Some(s)
+        case clob: java.sql.Clob => Some(
+          clob.getSubString(1, clob.length.toInt))
+      }
+
+      str match {
```
richdougherty: Is this ever going to match a `None` value? I think the definition of `str` means it will either give a `Some` value or throw a `MatchError`.

postgres/src/main/scala/package.scala
```diff
@@ -0,0 +1,127 @@
...
+import java.util.UUID
+
+import scala.util.control.NonFatal
+
+import play.api.libs.json.{
+  JsError, JsObject, Json, JsSuccess, JsValue, Reads, Writes }
+
+import org.postgresql.util.PGobject
+
+package object postgresql extends PGJson {
```
richdougherty: Have you thought about making some of this UUID/JSON conversion code generic so it can be used by other database backends? You could have a default implementation that serializes to a VARCHAR and then have the PostgreSQL implementation specialise that. One of the main benefits of this is that you could write an embedded unit test using H2.

build.sbt
```diff
@@ -29,6 +30,13 @@ lazy val anorm = project
       sourceManaged in Compile).map(m => Seq(GFA(m / "anorm"))),
     scalacOptions += "-Xlog-free-terms",
     binaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[MissingClassProblem]("anorm.MayErr"),
+      ProblemFilters.exclude[MissingClassProblem]("anorm.MayErr$"),
+      // was deprecated:
```
richdougherty: There are lots of miscellaneous implementation changes in anorm that change compatibility slightly. I know release compatibility is boring, but it's something we worry about. :)

Maybe we could bump the anorm version so that users can tell that there have been changes? Now that anorm is separated out from Play we can do that much more easily.

build.sbt
richdougherty: @cchantep can you explain this more? I see that play-json has a 2.6.x release for Scala 2.10, 2.11 and 2.12 so I think we should be able to use the same version for all of these versions of Scala.

core/src/main/scala/anorm/MayErr.scala
cchantep: Right it was "just" a limited wrapper, w/o all the `Either` semantics

build.sbt
cchantep: That's private changes, so it doesn't deserve a major bump for me.

postgres/src/main/scala/package.scala
cchantep: That's already the case for UUID, for Json I d prefer not to have such dependency in the core. I generic JSON module could be.

build.sbt
cchantep: As Anorm is not follow the same release cycle I don't see reason to bump.

postgres/src/main/scala/package.scala
cchantep:
```scala
case Some(js) => ...
case _ => // None
```

richdougherty: What I mean is that `str` only generates `Some` values. In other words you could probably change `str` to

```scala
@inline def str: String = value match {
  case o: PGobject => o.getValue)
  case s: String => s
  case clob: java.sql.Clob =>
  clob.getSubString(1, clob.length.toInt)
}
```

cchantep: I'd rather think the `case _ => None` is missing in `str` (good catch)

cchantep: Updated

cchantep:
Updated

cchantep:
Up

PR #156, opened 2017-08-27
Update dependencies and related refactoring

cchantep:


akka/src/test/scala/anorm/AkkaStreamSpec.scala
```diff
@@ -2,24 +2,29 @@ package anorm
...
-import org.specs2.concurrent.{ ExecutionEnv => EE }
+
+import org.specs2.concurrent.ExecutionEnv
+
 import acolyte.jdbc.AcolyteDSL.withQueryResult
 import acolyte.jdbc.RowLists.stringList
 import acolyte.jdbc.Implicits._
 
-class AkkaStreamSpec extends org.specs2.mutable.Specification {
+class AkkaStreamSpec(implicit ee: ExecutionEnv)
```
cchantep: Specs2 upgrade

build.sbt
```diff
@@ -2,21 +2,21 @@ import AnormGeneration.{ generateFunctionAdapter => GFA }
...
 
-lazy val acolyteVersion = "1.0.43-j7p"
+lazy val acolyteVersion = "1.0.46"
 lazy val acolyte = "org.eu.acolyte" %% "jdbc-scala" % acolyteVersion % Test
 
 lazy val `anorm-tokenizer` = project
   .in(file("tokenizer"))
   .enablePlugins(PlayLibrary, CopyPasteDetector)
-  .settings(scalariformSettings ++ Seq(
+  .settings(scalariformSettings(autoformat = true) ++ Seq(
```
cchantep: Scalariform & MiMa upgrade

core/src/main/scala/anorm/Macro.scala
```diff
@@ -551,11 +552,11 @@ object Macro {
 
   private def pretty(c: whitebox.Context)(parser: c.Tree): String =
     c.universe.show(parser).replaceAll("anorm.", "").
-      replaceAll("\\.\\$tilde", " ~ ").
-      replaceAll("\\(SqlParser([^(]+)\\(([^)]+)\\)\\)", "SqlParser$1($2)").
-      replaceAll("\\.\\$plus\\(([0-9]+)\\)", " + $1").
-      replaceAll("\\(([^ ]+) @ _\\)", "($1)").
-      replaceAll("\\$tilde", "~")
+      replaceAll(f"\\.\\$$tilde", " ~ ").
```
cchantep: Avoid `$` detected as interpolation

core/src/main/scala/anorm/Column.scala
```diff
@@ -477,40 +487,44 @@ object Column extends JodaColumn with JavaTimeColumn {
     def transf(a: Array[_], p: List[T]): Either[SqlRequestError, List[T]] =
       a.headOption match {
         case Some(r) => transformer(r, meta) match {
-          case Right(v) => transf(a.tail, p :+ v)
+          case Right(v) => transf(a.tail, v :: p)
```
richdougherty: I'm impressed with your attention to detail, fixing up this, and changing `Seq[T]()` to `Seq.empty[T]`!

core/src/main/scala/anorm/Row.scala
```diff
@@ -96,21 +97,22 @@ trait Row {
   def apply[B](position: Int)(implicit c: Column[B]): B =
     unsafeGet(SqlParser.get(position)(c))
 
+  @SuppressWarnings(Array("EitherGet")) // TODO: Safe alternative
   @inline def unsafeGet[T](rowparser: => RowParser[T]): T =
     (rowparser(this) match {
       case Success(v) => Right(v)
       case Error(err) => Left(err)
-    }).right.get // TODO: Safe alternative
+    }).right.get
```
richdougherty: I know this isn't related to this PR, but out of interest, why does this code convert a `Try` to an `Either`, but then assume the result is a `Right` by calling `.right.get`? This means that if the original `Try` was an `Error`, we lose the original error and now get the result of `LeftProjection.get`, which always throws an `NoSuchElementException`…

Instead of all this we could just do `rowparser(this).get`…

core/src/main/scala/anorm/ToSql.scala
```diff
@@ -122,4 +121,7 @@ object ToSql {
 
     sql._1.toString -> sql._2
   }
+
+  /** Fallback when no instance is available. */
+  def missing[A] = null
```
richdougherty: Do you want to put a type annotation on this so it's not inferred as `AnyRef`? E.g. `def missing[A]: ToSql[A] = null`.

core/src/main/scala/anorm/Column.scala
richdougherty: Or did this plugin help? https://github.com/sksamuel/sbt-scapegoat :)

cchantep: That's it

core/src/main/scala/anorm/Row.scala
cchantep: Good catch

core/src/main/scala/anorm/ToSql.scala
cchantep: That's a workaround as with explicit type this triggers a MiMa error:

```
... where it is anorm.ToSql rather than scala.runtime.Null#
```

Just added a `@deprecated` to advice not to use it outside from Anorm itself.

core/src/main/scala/anorm/Row.scala
cchantep: Refactored

richdougherty:
That change looks fine to me.

cchantep:
Thx @richdougherty

Issue #159, opened 2017-10-30
Publish anorm-postgres to Maven Central.


lavrov:

cchantep:
The issue is more "Publish the latest version" (all modules)

mkosovec:
Second this - currently debating building my own internal repository just to host the most recent version instead of having everyone publish locally.
The Postrges module is by far the most simple/clean way to bridge the gap between jsonb and jsValues in play.

btmcnellis:
Is there any update on when this might be published @cchantep?

cchantep:
I cannot answer

richdougherty:
I think @marcospereira was going to do this. Last I heard he was [asking on Gitter](https://gitter.im/playframework/contributors?at=5a32e978232e79134d4e553a) about the version number.

cchantep:
Thx for the feedback. There have several "ill-published" artifacts for Anorm (as 2.6.0-M1 or 3.0.0-M1). As there is no major API change/breakage on the existing modules (already published as 2.5.3), I don't see reason the master modules would not be published as 2.5.4. Best regards.

> See #162

gmethvin:
There are a number of compatibility issues, so I think we should bump the major version. There were a lot of binary issue filters added in #147 and that has not yet been released. I'm assuming we're following the Play epoch.major.minor where we avoid breaking changes in "minor" versions but can break things in "major" versions.

Another thing we could do is drop scala 2.11 support for 2.6.0, which will make it easier to add methods to traits (as you did in #147) without breaking bincompat. It would also be nice to add scala 2.13 support soon.

Given that the versioning scheme is different from that of Play, I wonder if we should consider going straight to 3.0.0. anorm-postgres is a new "feature" so it seems like enough justification. Some users have been a bit confused by the fact that play-json still uses similar versions to Play but is not released concurrently. This might be an opportunity to help avoid that kind of confusion for anorm.

gmethvin:
Not sure if you noticed but there are several public methods that are incompatible: https://github.com/playframework/anorm/pull/147/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR81. You added methods to `Column`: https://github.com/playframework/anorm/pull/147/files#diff-b16d230677d928f6fa2f6cb6eb1abab5R68.

cchantep:
Adding function is not the kind of evolution which deserve a major bump for me, as it doesn't raise backward compatibility issue for the existing user.

richdougherty:
[...]

PR #162, opened 2017-12-16
Prepare publication org.playframework.anorm 2.6.0

cchantep:
Fix compatibility, warnings and documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: scala
 jdk:
   - oraclejdk8
+  - oraclejdk9
 scala:
 - 2.12.4
-- 2.11.11
+- 2.11.12
 sudo: false
 cache:
   directories:
--- a/.travis_scripts/validate.sh
+++ b/.travis_scripts/validate.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 
-sbt ++$TRAVIS_SCALA_VERSION scalariformFormat test:scalariformFormat > /dev/null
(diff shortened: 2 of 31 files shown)


cchantep:
Group updated @richdougherty

mkurz:
@cchantep Why did you move to `org.playframework.anorm`? I can't find a answer in #159 as well. I guess that was because you wanted to have anorm published to maven central back then (while Play was still published to bintray)?
Actually anorm is the only artifact under `org.playframework.anorm`: https://repo1.maven.org/maven2/org/playframework/
Since last year, Play and all of it's sub-projects are now published to maven central as well: https://repo1.maven.org/maven2/com/typesafe/play/

Should we move anorm back to the `com.typesafe.play` groupId again? What do you think?

BTW: The `playframework` sonatype user now has publish permissions for `org.playframeework` as well (https://issues.sonatype.org/browse/OSSRH-36649), so we could also keep thing like they are now and use GH action for releases too (I will provide a pull request for that)

jroper:
@mkurz I just forwarded you an email thread. This was initiated by Lightbend, essentially, we wanted to give a wide range of community members the ability to publish artifacts, especially for Play subprojects, but we didn't want to grant them access to com.typesafe, since that's much bigger than just Play. So we created the new group id, org.playframework.

If we were going to change the group id of anything, I'd probably look at migrating Play to org.playframework, since the typesafe name doesn't mean much to anyone anymore. Of course, that's a big change that will break the entire ecosystem, and it will introduce a lot of problems with dependencies - any dependencies not upgraded to the new version will end up bringing com.typesafe.play.* artifacts into the classpath, which will provide the same classes as the new ones, and the effects will be very confusing. So, it might just be best not to do that.

PR #166, opened 2018-01-17
Add missing placeholder name in exception

vdebergue:
# Pull Request Checklist

* [x] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?
* [x] Have you added copyright headers to new files?
* [x] Have you checked that both Scala and Java APIs are updated?
* [x] Have you updated the documentation for both Scala and Java sections?
* [x] Have you added tests for any changed functionality?

## Purpose

Add a more helpful message in case of missing parameter in sql queries

## Background Context

When submiting a complex and long sql query with multiple parameters it's hard to know which parameter is missing. Here we add the name of the missing parameter in the exception to help the user.

For instance with a query:
```scala
SQL("SELECT * FROM test WHERE foo = {foo} AND bar = {bar}")
  .on("foo" -> 42)
  .execute()
```

We now get the message `Missing parameter value for 'bar' after: "AND bar = "`

### DIFF
--- a/core/src/main/scala/anorm/Anorm.scala
+++ b/core/src/main/scala/anorm/Anorm.scala
@@ -211,9 +211,9 @@ object Sql { // TODO: Rename to SQL
   }
 
   @SuppressWarnings(Array("IncorrectlyNamedExceptions"))
-  final class MissingParameter(after: String)
+  final class MissingParameter(after: String, placeholder: String)
     extends java.util.NoSuchElementException(
-      s"Missing parameter value after: $after") with NoStackTrace {}
+      s"Missing parameter value for '$placeholder' after: $after") with NoStackTrace {}
 
   object NoMorePlaceholder extends Exception("No more placeholder")
     with NoStackTrace {}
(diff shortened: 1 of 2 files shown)


vdebergue:
This seems to cause a binary incompatibility:

```
found 1 potential binary incompatibilities while checking against com.typesafe.play:anorm_2.11:2.5.0
 method this(java.lang.String)Unit in class anorm.Sql#MissingParameter does not have a correspondent in current
```

Any help on how this should be solved ?

cchantep:
Hey @vdebergue , running locally `./.travis_scripts/validate.sh` is reproducing the compat error?

vdebergue:
Yes it does

cchantep:
This is the [API change](https://github.com/playframework/anorm/pull/166/files#diff-32796d0372be367fa2dcf7e40d3c0323R214) breaking compat.

core/src/main/scala/anorm/Anorm.scala
```diff
@@ -211,9 +211,12 @@ object Sql { // TODO: Rename to SQL
   }
 
   @SuppressWarnings(Array("IncorrectlyNamedExceptions"))
-  final class MissingParameter(after: String)
+  final class MissingParameter(after: String, placeholder: String)
     extends java.util.NoSuchElementException(
-      s"Missing parameter value after: $after") with NoStackTrace {}
+      s"Missing parameter value for '$placeholder' after: $after") with NoStackTrace {
+    @deprecated("Create this exception while supplying the missing placeholder value", "2.6.1")
+    def this(after: String) = this(after, "")
```
cchantep: `"<unknown>"` ?

PR #182, opened 2018-09-17
'column not found' vs 'null value found'

gregor-i:
# Pull Request Checklist

* [x] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?
* [x] Have you added copyright headers to new files?
* [x] Have you added tests for any changed functionality?

## Purpose

I recently came across some unexpected behavior for nullable columns with the `?` operator of `RowParser`. It accepts non-existent columns and returns `None` as a result. I expected it to throw an exception because it could never return a value for this query. 

Changing the behavior of the `?` would introduce a breaking change of the API. So I added a new method on `RowParser` to explicitly handle null values in the result set.

## Background Context

The current behavior of the `?` can easily create bugs. For example a simple typo in the column name will stay undetected in all negative tests and will only be detected in a positive test. Using the new `nullable` operator instead will show this typo on the first execution of the query.

### DIFF
--- a/core/src/main/scala/anorm/SqlParser.scala
+++ b/core/src/main/scala/anorm/SqlParser.scala
@@ -621,6 +621,19 @@ trait RowParser[+A] extends (Row => SqlResult[A]) { parent =>
     }
   }
 
+  /**
+   * Returns a row parser for nullable column,
+   * that will turn null into None.
+   */
+  def nullable: RowParser[Option[A]] = RowParser {
+    parent(_) match {
+      case Success(a) => Success(Some(a))
+      case Error(UnexpectedNullableFound(_)) =>
+        Success(None)
+      case e @ Error(_) => e
+    }
+  }
+
   /** Alias for [[flatMap]] */
   def >>[B](f: A => RowParser[B]): RowParser[B] = flatMap(f)
 
--- a/core/src/test/scala/anorm/AnormSpec.scala
+++ b/core/src/test/scala/anorm/AnormSpec.scala
@@ -142,6 +142,53 @@ class AnormSpec extends Specification with H2Database with AnormTest {
         }
     }
 
+    "handle nullable columns" >> {
+      "return instance with defined option" in withQueryResult(rowList2(
+        classOf[Int] -> "id", classOf[String] -> "val") :+ (2, "str")) {
+        implicit c =>
+
+          SQL("SELECT * FROM test").as(
+            SqlParser.int("id") ~ SqlParser.str("val").nullable map {
+              case id ~ v => (id -> v)
+            } single) aka "mapped data" must_== (2 -> Some("str"))
+
+      }
+
+      "return instance with None for null value found" in withQueryResult(
+        rowList2(classOf[Long] -> "id", classOf[String] -> "val") :+ (2, null)) { implicit c =>
+
+          SQL("SELECT * FROM test").as(
+            SqlParser.long("id") ~ SqlParser.str("val").nullable map {
+              case id ~ v => (id -> v)
+            } single) aka "mapped data" must_== (2 -> None)
+
+        }
+
+      "throw exception when column dosen't exist" in withQueryResult(
+        rowList1(classOf[Long] -> "id") :+ 123l) { implicit c =>
+
+          SQL("SELECT * FROM test").as(
+            SqlParser.long("id") ~ SqlParser.str("val").nullable map {
+              case id ~ v => (id -> v)
+            } single) aka "parser" must throwA[Exception].like {
+              case e: Exception => e.getMessage aka "error" must startWith(
+                "'val' not found, available columns: id, id")
+            }
+        }
+
+      "throw exception when type doesn't match" in withQueryResult(
+        fooBarTable :+ (1l, "str", 3)) { implicit c =>
+
+          SQL("SELECT * FROM test").as(
+            SqlParser.long("id") ~ SqlParser.int("foo").nullable map {
+              case id ~ v => (id -> v)
+            } single) aka "parser" must throwA[Exception].like {
+              case e: Exception => e.getMessage aka "error" must startWith(
+                "TypeDoesNotMatch(Cannot convert str:")
+            }
+        }
+    }
+
     "throw exception when type doesn't match" in withQueryResult("str") {
       implicit c =>
         SQL("SELECT * FROM test").as(scalar[Int].single).


core/src/main/scala/anorm/SqlParser.scala
```diff
@@ -621,6 +621,19 @@ trait RowParser[+A] extends (Row => SqlResult[A]) { parent =>
     }
   }
 
+  /**
+   * Returns a row parser for nullable column,
+   * that will turn null into None.
+   */
+  def nullable: RowParser[Option[A]] = RowParser {
```
cchantep: There are already `.?` for column and `.singleOpt`, so what ?

gregor-i: Ok, let me give you an example:
```scala
SQL"SELECT 1 AS test".as(SqlParser.int("test").?.single)
```

But you make a mistake and do this instead:
```scala
SQL"SELECT 1 AS test".as(SqlParser.int("TYPO").?.single)
```
Then you may not detect your mistake, because `None` might be an acceptable response.

But if you use `.nullable` instead of `.?` you query will throw an error and you will be informed about your mistake.
```scala
SQL"SELECT 1 AS test".as(SqlParser.int("TYPO").nullable.single)
```


This is a reduced example. In productive code I've got a query with 17 named fields and changed the name of a field without changing the parameter in the parser. This lead to unexpected behavior.

cchantep: Which mistake? If you want to require the column and have an error, rather use scalar/not optional column parser. The "expected behavior" is quite unclear to me.

gregor-i: The mistake is that I'm referencing a not existing column.

The `.?` method catches `UnexpectedNullableFound` and `ColumnNotFound`. From my point of view these are 2 different kinds of problems. `ColumnNotFound` is a problem with the query itself. `UnexpectedNullableFound` on the other hand might not be any problem at all. 

So to differentiate between these 2 kinds of problems, I wanted a different method which doesn't catch `ColumnNotFound` and only catches `UnexpectedNullableFound`.

cchantep: I don't think these assumption correspond to underlying JDBC, neither such distinction is found in similar context, such as JSON field.

gregor-i: JDBC and even `SqlRequestError` differentiates between these types of problem. See [ColumnNotFound](https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/SqlRequestError.scala#L54) and [UnexpectedNullableFound](https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/SqlRequestError.scala#L63).

And so does json. For example in javascript `undefined === null` will be evaluated to `false`.

cchantep: JSON is not JavaScript, and doesn't require distinction between missing and null, moreover null logic is not idiomatic in Scala.

Maybe the current error kinds can be simpler, considering the suggested nullable combinator, I'm not convince about its utility.

gregor-i: but json does differentiate between null and undefined. Even play-json does so:

```
Starting scala interpreter...
Welcome to Scala 2.12.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_77).
Type in expressions for evaluation. Or try :help.

scala> import play.api.libs.json.{JsObject, Json}
import play.api.libs.json.{JsObject, Json}

scala> val n = Json.parse("""{"x": null}""")
n: play.api.libs.json.JsValue = {"x":null}

scala> val u = Json.parse("""{}""")
u: play.api.libs.json.JsValue = {}

scala> n.as[JsObject].value.size
res0: Int = 1

scala> u.as[JsObject].value.size
res1: Int = 0
```

And even if null is not idiomatic in scala, it is in SQL.

Please don't merge these error types. There are good reasons to distinguish these errors ... 

I needed this utility, I implemented in my project, and it helped me a lot. That is enough reason for me to post this PR. If you are not convinced, just reject the PR.

cchantep: `UnexpectedNullableFound` is only raised by instances of the `Column` typeclass, whereas `ColumnNotFound` is the same at row level (`RowParser`) with the row meta (column names) as context (which is not known at isolated column level).

I would rather update `.scalar` so that it normalized `UnexpectedNullableFound` as `ColumnNotFound` before it bubbles up to the row level.

The suggested `.nullable` combinator doesn't make sense to me.

cchantep: Considering the following `ResultSet` from table `t`:

Foo|Bar
---|---
 A | 2

## Case.1

```scala
// No 'Lorem' column
SQL"SELECT * FROM t".as(int("Lorem").?.single)
```

Returns `None` (as Play JSON, see previous example).

If you want to enforce the naming of your mapping in such case, I would rather suggest to explicit the projection columns (instead of `*`), and let the DB do this job.

```
val ColName = "Lorem"
SQL"SELECT #${ColName} FROM t".as(int(ColName).?.single)
```

Enforce the consistency between the statement and the mappings, and this lets the DB complains that the `${ColName="Lorem"}` cannot be resolved in the table `t`, so it will fail at query execution, before the mapping (`.as`), and you can log it if you want as any execution error.

## Case.2

```scala
// No 'Lorem' column
SQL"SELECT Foo FROM t".as(int("Lorem").?.single)
```

Even with explicit projection, the consistency is not enforce between the statement and the mapping. It can be done as in Case.1.

```
val ColName = "Lorem"
SQL"SELECT #${ColName} FROM t".as(int(ColName).?.single)
```

As in Case.1, it will make sure an execution error is raised (benefiting from the DB query validation).

cchantep: The proposed `.nullable` doesn't give ensure about the column naming consistency, which is the addressed by the extra code.

My point is for this use case, there are other (possibly better solution, but that's another discussion), which don't require to increase the API complexity (as Anorm aims to keep it simple).

gregor-i: It does ensure that the column exist. That's the whole point. It behaves the same way as a mandatory column:

```scala
// nullable:
    SQL"SELECT 'lorem' AS lorem".as(str("lorem").nullable.single) shouldBe Some("lorem")
    SQL"SELECT null AS lorem".as(str("lorem").nullable.single) shouldBe None
    assertThrows[anorm.AnormException](SQL"SELECT null AS foo".as(str("lorem").nullable.single))

// ?:
    SQL"SELECT 'lorem' AS lorem".as(str("lorem").?.single) shouldBe Some("lorem")
    SQL"SELECT null AS lorem".as(str("lorem").?.single) shouldBe None
    SQL"SELECT null AS foo".as(str("lorem").?.single) shouldBe None

// mandatory:
    SQL"SELECT 'lorem' AS lorem".as(str("lorem").single) shouldBe "lorem"
    assertThrows[anorm.AnormException](SQL"SELECT null AS lorem".as(str("lorem").single))
```

cchantep:
Close it for now. Feel free to update the documentation to explicit the `.?` edge cases discussed there.

Issue #325, opened 2021-01-06
Default Column[Instant] instance loose nano seconds precision


FXHibon:
### Anorm Version
- 2.6.8

### Operating System
- MacOS 11.1

### JDK

tested on both:
- OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_272-b10)
- OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9+11)

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

1. insertion of a **java.time.Instant** value with nano seconds precision in a database table, using the standard **instantToStatement: ToStatement[Instant]** (https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/ToStatementMisc.scala#L727)
2. read the inserted row, using the **columnToInstant: Column[Instant]** to build a **RowParser[Instant]** (https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/Column.scala#L713)
3. I expect the value to be exactly the same as the value inserted step 1

### Actual Behavior

Please provide a description of what actually happens, working from the same starting point.

1. insert the row
2. read the row
3. the value read from the database is truncated to miliseconds, loosing the nano seconds precision. This is not what I expected.

### Reproducible Test Case

Here a complete and minimal example that illustrate the problem.

https://github.com/FXHibon/play-anorm-nano-precision

The failings test can be accessed here => https://github.com/FXHibon/play-anorm-nano-precision/blob/master/test/repositories/InstantRepositoryTest.scala#L39

### What else?

I noticed this problem when I switched my app from JDK 8 to JDK 11
Using JDK 8, **Instant** instances generated with `java.time.Instant.now` are **not** nano second precision
Using JDK 11, same code will generate instances with nano seconds precision
See screen below to illustrate this
![image](https://user-images.githubusercontent.com/5735335/103773870-03f36d00-502c-11eb-9536-0b2324c01b8e.png)

Consequently, the reported issue is less likely to happen using JDK 8

I will quickly provide a pull request to fix the default **Column[Instant]**, so the nano seconds are not lost during the reading of sql result

PR #326, opened 2021-01-06
FIX - Issue 325 - Column[Instant] loose nano seconds precision

FXHibon:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)? yes
* [x] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)? yes
* [x] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)? yes
* [x] Have you [squashed your commits](https://www.playframework.com/documentation/2.4.x/WorkingWithGit#Squashing-commits)? yes
* [x] Have you added copyright headers to new files? not relevant
* [x] Have you checked that both Scala and Java APIs are updated? no api changes
* [x] Have you updated the documentation for both Scala and Java sections? no changes
* [x] Have you added tests for any changed functionality? yes

## Fixes

Fixes [#325](https://github.com/playframework/anorm/issues/325)

## Purpose

Fix the default instance of **Column[Instant]** in the **JavaTimeColumn** trait, so we don't loose the nano second precision during the SQL result reads.

## Background Context

When reading an SQL result, **TIMESTAMP[TZ]** from database is mapped on the **java.sql.Timestamp** type. Therefore, the **Column[Instant]** should be able to transform a given **java.sql.Timestamp** into an **java.time.Instant**

As **java.sql.Timestamp** inherits **java.util.Date**, the first branch of following pattern-matching will be selected

```scala
   value match {
      case date: java.util.Date => Right(Instant ofEpochMilli date.getTime)
      case time: Long => Right(Instant ofEpochMilli time)
      case TimestampWrapper1(ts) => Ts(ts)(Instant ofEpochMilli _.getTime)
      case TimestampWrapper2(ts) => Ts(ts)(Instant ofEpochMilli _.getTime)
      case _ => Left(TypeDoesNotMatch(s"Cannot convert $value: ${className(value)} to Java8 Instant for column $qualified"))
    }
```

https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/Column.scala#L715

The problem is that **java.util.Date** is precised at the miliseconds level, and not the nano seconds level. Dealing with a **java.sql.Timestamp** as as **java.util.Date** will loose this precision, as we can understand immediatly with the following code:
```scala
Instant ofEpochMilli date.getTime
```
(**getTime** returns EPOCH miliseconds, as explained in javadoc)

I wonder why the pattern-matching wasn't using directly the **java.sql.Timestamp** type? Maybe is there an hidden reason I couldn't understand?

Adding a line (in first position) with a pattern on **java.sql.Timestamp** seems to be ok, especially since **Timestamp** can be directly converted to **Instant** with the method **.toInstant**.

The 2 lines that matches **TimestampWrapper1** and **TimestampWrapper2** would loose in the same way, because it build the Instant using only the miliseconds.

I took this opportunty to use **ts.toInstant** here to, to avoid silent data loss behavior.

Actually, I'm not sure to understand when these **TimestampWrapper** become useful, but I might not have the full picture I my mind.

## References

The issue is explained in details here https://github.com/playframework/anorm/issues/325

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,7 @@ import com.typesafe.tools.mima.plugin.MimaKeys.{
 ThisBuild / scalaVersion := "2.12.12"
 
 ThisBuild / crossScalaVersions := Seq(
-  "2.11.12", (scalaVersion in ThisBuild).value, "2.13.3")
+  "2.11.12", (scalaVersion in ThisBuild).value, "2.13.4")
 
 // Scalafix
 inThisBuild(
--- a/core/src/main/scala/anorm/Column.scala
+++ b/core/src/main/scala/anorm/Column.scala
@@ -713,10 +713,11 @@ sealed trait JavaTimeColumn {
   implicit val columnToInstant: Column[Instant] = nonNull { (value, meta) =>
     val MetaDataItem(qualified, _, _) = meta
     value match {
+      case ts: java.sql.Timestamp => Ts(ts)(_.toInstant)
       case date: java.util.Date => Right(Instant ofEpochMilli date.getTime)
       case time: Long => Right(Instant ofEpochMilli time)
-      case TimestampWrapper1(ts) => Ts(ts)(Instant ofEpochMilli _.getTime)
-      case TimestampWrapper2(ts) => Ts(ts)(Instant ofEpochMilli _.getTime)
+      case TimestampWrapper1(ts) => Ts(ts)(_.toInstant)
+      case TimestampWrapper2(ts) => Ts(ts)(_.toInstant)
       case _ => Left(TypeDoesNotMatch(s"Cannot convert $value: ${className(value)} to Java8 Instant for column $qualified"))
     }
   }
--- a/core/src/test/scala/anorm/JavaTimeSpec.scala
+++ b/core/src/test/scala/anorm/JavaTimeSpec.scala
@@ -1,5 +1,6 @@
 package anorm
 
+import java.sql.Timestamp
 import java.time.{ Instant, LocalDate, LocalDateTime, ZoneId, ZonedDateTime }
 
 import acolyte.jdbc.AcolyteDSL._
@@ -34,6 +35,16 @@ class JavaTimeColumnSpec extends Specification {
         SQL("SELECT ts").as(scalar[Instant].single).
           aka("parsed instant") must_=== instant
       }
+
+    "be parsed from timestamp with nano precision" in {
+      val instantWithNanoPrecision = Instant.parse("2021-01-06T08:45:26.441477Z")
+
+      withQueryResult(
+        timestampList :+ Timestamp.from(instantWithNanoPrecision)) { implicit con =>
+          SQL("SELECT ts").as(scalar[Instant].single).
+            aka("parsed instant") must_=== instantWithNanoPrecision
+        }
+    }
 
     "be parsed from numeric time" in withQueryResult(longList :+ time) {
       implicit con =>
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target
 logs
 project/project
+.bsp


core/src/main/scala/anorm/Column.scala
```diff
@@ -713,10 +713,11 @@ sealed trait JavaTimeColumn {
   implicit val columnToInstant: Column[Instant] = nonNull { (value, meta) =>
     val MetaDataItem(qualified, _, _) = meta
     value match {
+      case ts: java.sql.Timestamp => Right(ts.toInstant)
```
cchantep, line 716: That's handled by `case TimestampWrapper1(ts)` so why this duplicate?

core/src/test/scala/anorm/JavaTimeSpec.scala
```diff
@@ -35,6 +36,17 @@ class JavaTimeColumnSpec extends Specification {
           aka("parsed instant") must_=== instant
       }
 
+    "be parsed from timestamp with nano precision" in {
+
```
cchantep, line 40:
```suggestion
```

core/src/main/scala/anorm/Column.scala
```diff
@@ -713,10 +713,11 @@ sealed trait JavaTimeColumn {
...
       case date: java.util.Date => Right(Instant ofEpochMilli date.getTime)
       case time: Long => Right(Instant ofEpochMilli time)
-      case TimestampWrapper1(ts) => Ts(ts)(Instant ofEpochMilli _.getTime)
```
cchantep, line 718: Why not `Ts(ts)(_.toInstant)` to keep it consistent?

core/src/test/scala/anorm/JavaTimeSpec.scala
```diff
@@ -35,6 +36,17 @@ class JavaTimeColumnSpec extends Specification {
...
+      val instantWithNanoPrecision = Instant.parse("2021-01-06T08:45:26.441477Z")
+
+      withQueryResult(
+        timestampList :+ Timestamp.from(instantWithNanoPrecision)) { implicit con =>
+          SQL("SELECT ts").as(scalar[Instant].single).
+            aka("parsed instant") must_=== instantWithNanoPrecision
```
cchantep, line 46: For history, this test without code change:

```
[error]   x be parsed from timestamp with nano precision
[error]    parsed instant '2021-01-06T08:45:26.441Z != 2021-01-06T08:45:26.441477Z' (JavaTimeSpec.scala:46)
[error] Actual:   ...6.441[]Z
[error] Expected: ...6.441[477]Z
```

core/src/main/scala/anorm/Column.scala
FXHibon: **TimestampWrapper1** and **TimestampWrapper2** check the existence of respectively fields **getTimestamp: java.sql.Timestamp** or **timestampValue: java.sql.Timestamp**

but **java.sql.Timestamp** has no such fields

cf screen of my IDE (which might not be the best proof, I agree)


image


image



I've just checked the behavior by printing the class name of the value in each branch, after commenting my added line:

```scala
value match {
      //      case ts: java.sql.Timestamp => Right(ts.toInstant)
      case time: Long => Right(Instant ofEpochMilli time)
      case TimestampWrapper1(ts) =>
        println("matching TimestampWrapper1 " + className(value))
        Right(ts.toInstant)
      case TimestampWrapper2(ts) =>
        println("matching TimestampWrapper2 " + className(value))
        Right(ts.toInstant)
      case date: java.util.Date =>
        println("matching date " + className(value))
        Right(Instant ofEpochMilli date.getTime)
      case _ => Left(TypeDoesNotMatch(s"Cannot convert $value: ${className(value)} to Java8 Instant for column $qualified"))
    }
```

and I can see

> matching date java.sql.Timestamp

Therefore, as it, **TimestampWrapper**s cannot match **java.sql.Timestamp**, which might be confusing

build.sbt
```diff
@@ -9,7 +9,7 @@ import com.typesafe.tools.mima.plugin.MimaKeys.{
 ThisBuild / scalaVersion := "2.12.12"
 
 ThisBuild / crossScalaVersions := Seq(
-  "2.11.12", (scalaVersion in ThisBuild).value, "2.13.3")
+  "2.11.12", (scalaVersion in ThisBuild).value, "2.13.4")
```
FXHibon, line 12: not sure if this PR is the right time to upgrade this

I've always the reflex of upgrading patch semver when I can, let me now if it should not be done here

core/src/main/scala/anorm/Column.scala
FXHibon: this is a good point, I'll correct this

cchantep (approved):
Thanks for your contribution

FXHibon:
hello @cchantep 
Would you be able to release a version 2.6.9 of anorm that would include this PR?

cchantep:
Rather 2.6.10

cchantep:
https://search.maven.org/artifact/org.playframework.anorm/anorm_2.12/2.6.9/jar

PR #333, opened 2021-02-09
columnToLocalDateTime should accept LocalDateTime

### LINKED ISSUE #329: java.time.LocalDateTime cannot be converted to Java8 LocalDateTime
Anorm has become unable to parse a `LocalDateTime` from a MySQL database. This has worked wonderfully for me until just recently - a few days back. I suspect some other library that has recently been updated may be at play (no pun intended), but I haven't been able to pinpoint the source of the change. However, the error seems to point to something that I suspect Anorm _should_ handle - accepting a `java.time.LocalDateTime` as a `LocalDateTime` - `TypeDoesNotMatch(Cannot convert 2011-03-13T02:49:10: java.time.LocalDateTime to Java8 ZonedDateTime for column ColumnName`.

I'm very happy to provide more detail around my specific failure case if it helps to narrow down how to replicate, or what the issue might be. For now, I've put in a minimal amount of detail in case the problem or fix seems obvious to someone else.

### Anorm Version (2.5.x / etc)

2.6.8

Ubuntu 18.04 (Kernel version 4.15.0-132-generic)

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

OpenJDK 15.0.1 (build 15.0.1+9-18)

### Library Dependencies

* `mysql-connector-java` version 8.0.23
* `play-json` version 1.7.2
* `play-framework` version 2.8.2
* Scala 2.13.4

### Expected Behavior

1. Create MySQL table: 

```
CREATE TABLE examples (
  id varchar(64) NOT NULL,
  local_datetime datetime NOT NULL,
  time_zone varchar(64) NOT NULL,
  data varchar(256) DEFAULT NULL,
  PRIMARY KEY (`id`)
);
```

2. Insert row into table
```
INSERT INTO `examples` VALUES ('5b7c9d70-5e51-4ea0-bd0a-75457d7737ca','2011-03-13 02:49:10','Africa/Dar_es_Salaam','other data');
```

3. Create parser:
```
val EXAMPLE_PARSER: RowParser[Example]  = {
  get[UUID]("id") ~
  get[LocalDateTime]("local_datetime") ~
  get[String]("time_zone") ~
  get[Option[String]]("data") map { case id ~ localDateTime ~ timeZone ~ data =>
    Example(id, localDateTime.atZone(ZoneId.of(timeZone)), data)
  }
}
```

4. Query for data, and format using parser:
```
SQL"""SELECT * FROM examples""".as(ExampleDAO.EXAMPLE_PARSER.*)
```

### Actual Behavior

On step 4 above, Anorm reports an error in deserializing the `Example` record. It reports the exception `anorm.AnormException: TypeDoesNotMatch(Cannot convert 2011-03-13T02:49:10: java.time.LocalDateTime to Java8 ZonedDateTime for column ColumnName(examples.local_datetime,Some(local_datetime)))`

### Reproducible Test Case

I'm not quite sure what additional test needs to be added to `JavaTimeSpec` to show this failure.

pschichtel:
# Pull Request Checklist

* [X] ~~Have you added copyright headers to new files?~~
* [X] ~~Have you checked that both Scala and Java APIs are updated?~~
* [X] ~~Have you updated the documentation for both Scala and Java sections?~~
* [X] Have you added tests for any changed functionality?

## Fixes

Fixes #329

## Purpose

The latest MySQL JDBC driver returns LocalDateTime instances for datetime columns, however the existing `Column[LocalDateTime]` definition does not support LocalDateTime objects in its pattern match.

## Background Context

I changed to the column definition so that it returns the LocalDateTime instance directly without first converting epoch millis and back again. Other than that, this change was fairly straight forward.

### DIFF
--- a/core/src/main/scala/anorm/Column.scala
+++ b/core/src/main/scala/anorm/Column.scala
@@ -722,7 +722,7 @@ sealed trait JavaTimeColumn {
     }
   }
 
-  private def temporalColumn[T](epoch: Long => T, description: String): Column[T] = nonNull { (value, meta) =>
+  private def temporalValueTo[T](epoch: Long => T, description: String)(value: Any, meta: MetaDataItem): Either[SqlRequestError, T] = {
     val MetaDataItem(qualified, _, _) = meta
     value match {
       case date: java.util.Date => Right(epoch(date.getTime))
@@ -732,6 +732,9 @@ sealed trait JavaTimeColumn {
       case _ => Left(TypeDoesNotMatch(s"Cannot convert $value: ${className(value)} to $description for column $qualified"))
     }
   }
+
+  private def temporalColumn[T](epoch: Long => T, description: String): Column[T] =
+    nonNull(temporalValueTo(epoch, description))
 
   /**
    * Parses column as Java8 local date/time.
@@ -748,10 +751,17 @@ sealed trait JavaTimeColumn {
    *     as(SqlParser.scalar[LocalDateTime].single)
    * }}}
    */
-  implicit val columnToLocalDateTime: Column[LocalDateTime] =
-    temporalColumn[LocalDateTime]({ ts: Long =>
+  implicit val columnToLocalDateTime: Column[LocalDateTime] = nonNull { (value, meta) =>
+
+    def millisToLocalDateTime(ts: Long) =
       LocalDateTime.ofInstant(Instant.ofEpochMilli(ts), ZoneId.systemDefault)
-    }, "Java8 LocalDateTime")
+
+    value match {
+      case localDateTime: LocalDateTime => Right(localDateTime)
+      case _ =>
+        temporalValueTo[LocalDateTime](millisToLocalDateTime, "Java8 LocalDateTime")(value, meta)
+    }
+  }
 
   /**
    * Parses column as Java8 local date.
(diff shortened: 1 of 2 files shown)


core/src/main/scala/anorm/Column.scala
```diff
@@ -748,10 +751,17 @@ sealed trait JavaTimeColumn {
...
-  implicit val columnToLocalDateTime: Column[LocalDateTime] =
-    temporalColumn[LocalDateTime]({ ts: Long =>
+  implicit val columnToLocalDateTime: Column[LocalDateTime] = nonNull { (value, meta) =>
+
+    def millisToLocalDateTime(ts: Long) =
       LocalDateTime.ofInstant(Instant.ofEpochMilli(ts), ZoneId.systemDefault)
-    }, "Java8 LocalDateTime")
+
+    value match {
+      case localDateTime: LocalDateTime => Right(localDateTime)
```
cchantep, line 760: How a `LocalDateTime` could be there in JDBC ?

pschichtel: https://dev.mysql.com/doc/relnotes/connector-j/8.0/en/news-8-0-23.html

From the important changes:

> Also, with the implementation of the new mechanism, a getObject(columnIndex) call on a DATETIME column returns a LocalDateTime object now instead of a String. To receive a String like before, use getObject(columnIndex, String.class) instead. 

I haven't read any specs, but I assume getObject(int) is free to return any object.

pschichtel: I guess you could freely interpret this to allow for that:

> This method may also be used to read database-specific abstract data types.

\- https://docs.oracle.com/javase/8/docs/api/java/sql/ResultSet.html#getObject-int-

PR #409, opened 2021-11-30
FIX - Issue #408 - Column[ZonedDateTime] loses nanoseconds precision

### LINKED ISSUE #325: Default Column[Instant] instance loose nano seconds precision

### LINKED ISSUE #408: Default Column[ZonedDateTime] instance lose nanoseconds precision
### Anorm Version

- 2.6.10

### Operating System

- MacOS 11.1

### JDK

- OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9+11)

### Library Dependencies

- PostgreSQL 9.6.12
- Driver "org.postgresql:postgresql:42.3.1"

### Expected Behavior

1. Insertion of a `java.time.ZonedDateTime` value with nano seconds precision in a database table, using the standard `zonedDateTimeToStatement: ToStatement[ZonedDateTime]` (https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/ToStatementMisc.scala#L779)
2. Read the inserted row, using the `columnToZonedDateTime: Column[ZonedDateTime]` to build a RowParser[ZonedDateTime] (https://github.com/playframework/anorm/blob/master/core/src/main/scala/anorm/Column.scala#L814)
3. The returned value should be the same as the value inserted step 1

### Actual Behavior

1. Insertion of a `java.time.ZonedDateTime` value with nano seconds precision
2. Read the inserted row
3. The returned value is truncated to miliseconds, loosing the nano seconds precision

This issue is very similar to the issue #325 

### Reproducible Test Case

https://github.com/rrramiro/play-anorm-nano-precision

rrramiro:
# Pull Request Checklist

* [X] Have you added copyright headers to new files?
* [X] Have you checked that both Scala and Java APIs are updated?
* [X] Have you updated the documentation for both Scala and Java sections?
* [X] Have you added tests for any changed functionality?

## Fixes

Fixes #408

## Purpose

Fix the default instance of `Column[ZonedDateTime]` in the `JavaTimeColumn` trait, so it doesn't lose the nanosecond precision during the SQL result reads.

## Background Context

When reading an SQL result, `TIMESTAMP[TZ]` from database is mapped on the `java.sql.Timestamp` type. This type extends from `java.util.Date` but has a nanosecond precision. In current implementation the conversion is handle by `temporalValueTo` function:
```scala
  private def temporalValueTo[T](epoch: Long => T, description: String)(value: Any, meta: MetaDataItem): Either[SqlRequestError, T] = {
    value match {
      case date: java.util.Date => Right(epoch(date.getTime)) // Timestamp conversion
      case time: Long => Right(epoch(time))
      case TimestampWrapper1(ts) => Ts(ts)(t => epoch(t.getTime))
      case TimestampWrapper2(ts) => Ts(ts)(t => epoch(t.getTime))
      case _ => Left(TypeDoesNotMatch(s"Cannot convert $value: ${className(value)} to $description for column $qualified"))
    }
  }
```
`java.sql.Timestamp` is matched as a `java.util.Date` and `date.getTime` only provides milliseconds.

There was already a fix for the `Instant` type:

```scala
  implicit val columnToInstant: Column[Instant] = nonNull { (value, meta) =>

    value match {
      case date: LocalDateTime => Right(date.toInstant(ZoneOffset.UTC))
      case ts: java.sql.Timestamp => Ts(ts)(_.toInstant) //Timestamp conversion with nanoseconds
      case date: java.util.Date => Right(Instant ofEpochMilli date.getTime)
      case time: Long => Right(Instant ofEpochMilli time)
      case TimestampWrapper1(ts) => Ts(ts)(_.toInstant)
      case TimestampWrapper2(ts) => Ts(ts)(_.toInstant)
      case _ => Left(TypeDoesNotMatch(s"Cannot convert $value: ${className(value)} to Java8 Instant for column $qualified"))
    }
  }
```

My suggestion is to reuse this fix and to use `Instant` as the intermediate type for date/time conversion instead of `Long`.
So that all conversion will have to provide a function `Instant => T` when it was `Long => T`.

## References

`Column[Instant]`  fix  #325 and PR #326

### DIFF
--- a/core/src/test/scala/anorm/JavaTimeSpec.scala
+++ b/core/src/test/scala/anorm/JavaTimeSpec.scala
@@ -188,6 +188,15 @@ class JavaTimeColumnSpec extends Specification {
           aka("parsed zoned date/time") must_=== date
       }
 
+    "be parsed from timestamp with nano precision" in {
+      val instantWithNanoPrecision = Instant.parse("2021-01-06T08:45:26.441477Z")
+      withQueryResult(
+        timestampList :+ java.sql.Timestamp.from(instantWithNanoPrecision)) { implicit con =>
+          SQL("SELECT ts").as(scalar[ZonedDateTime].single).
+            aka("parsed zoned date/time") must_=== ZonedDateTime.ofInstant(instantWithNanoPrecision, ZoneId.systemDefault)
+        }
+    }
+
     "be parsed from numeric time" in withQueryResult(longList :+ time) {
       implicit con =>
         SQL("SELECT time").as(scalar[ZonedDateTime].single).
--- a/core/src/main/scala/anorm/Column.scala
+++ b/core/src/main/scala/anorm/Column.scala
@@ -718,32 +718,22 @@ sealed trait JavaTimeColumn {
    *   SQL("SELECT last_mod FROM tbl").as(scalar[Instant].single)
    * }}}
    */
-  implicit val columnToInstant: Column[Instant] = nonNull { (value, meta) =>
-    val MetaDataItem(qualified, _, _) = meta
+  implicit val columnToInstant: Column[Instant] = temporalColumn(identity[Instant], "Java8 Instant")
 
-    value match {
-      case date: LocalDateTime => Right(date.toInstant(ZoneOffset.UTC))
-      case ts: java.sql.Timestamp => Ts(ts)(_.toInstant)
-      case date: java.util.Date => Right(Instant ofEpochMilli date.getTime)
-      case time: Long => Right(Instant ofEpochMilli time)
-      case TimestampWrapper1(ts) => Ts(ts)(_.toInstant)
-      case TimestampWrapper2(ts) => Ts(ts)(_.toInstant)
-      case _ => Left(TypeDoesNotMatch(s"Cannot convert $value: ${className(value)} to Java8 Instant for column $qualified"))
-    }
-  }
-
-  private def temporalValueTo[T](epoch: Long => T, description: String)(value: Any, meta: MetaDataItem): Either[SqlRequestError, T] = {
+  private def temporalValueTo[T](epoch: Instant => T, description: String)(value: Any, meta: MetaDataItem): Either[SqlRequestError, T] = {
     val MetaDataItem(qualified, _, _) = meta
     value match {
-      case date: java.util.Date => Right(epoch(date.getTime))
-      case time: Long => Right(epoch(time))
-      case TimestampWrapper1(ts) => Ts(ts)(t => epoch(t.getTime))
-      case TimestampWrapper2(ts) => Ts(ts)(t => epoch(t.getTime))
+      case date: LocalDateTime => Right(epoch(date.toInstant(ZoneOffset.UTC)))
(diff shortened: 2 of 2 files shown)


lightbend-cla-validator:
At least one pull request committer is not linked to a user. See https://help.github.com/en/articles/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

rrramiro:
Thanks for the review @cchantep. You're right no need to use `Instant` for `LocalDateTime`

core/src/test/scala/anorm/JavaTimeSpec.scala
```diff
@@ -188,6 +188,15 @@ class JavaTimeColumnSpec extends Specification {
           aka("parsed zoned date/time") must_=== date
       }
 
+    "be parsed from timestamp with nano precision" in {
+      val instantWithNanoPrecision = Instant.parse("2021-01-06T08:45:26.441477Z")
```
markarasev, line 192: Is it me or the tested instant has a millisecond precision instead of nanosecond?

rrramiro: The tested instant has `441477` **micro**seconds. The issue was happening for all the presisions under millisecond. So the test is not wrong, just the naming.

PR #433, opened 2022-02-08
Provide a Column to convert bytes and shorts to bools 

### LINKED ISSUE #432: Boolean columns broken with MariaDB driver v3
After upgrading the MariaDB JDBC driver from version 2 to version 3, boolean columns (SQL type `BOOLEAN`, Scala type `Boolean`) fail to be parsed.

I have not dug deeply into the root cause, but it could be that metadata reports them as `Byte` instead. I guess that makes sense to some degree, as MariaDB booleans are really synonyms for `tinyint(1)` according to [the docs](https://mariadb.com/kb/en/boolean/). The French docs even mention that SQL-compliant boolean implementations will come in a future version, but for some reason this mention is absent in the English docs.

In any case, it seems acceptable to handle the *byte -> boolean* case as a match in `columnToBoolean`. I realize that there is documentation explaining how to do it with [custom mappings](https://www.playframework.com/documentation/2.6.x/ScalaAnorm#JDBC-mappings), but it seems so "standard" that I believe it should be part of the default mapping. Case in point, the symmetric *boolean -> byte* case is already implemented in `columnToByte`.

### Anorm Version

2.6.10

### JDK

Java 11

### Library Dependencies

"org.mariadb.jdbc" % "mariadb-java-client" % "3.0.3"

### Expected Behavior

`RowParser[Boolean]` works on column that are declared in SQL script as type `BOOLEAN`

### Actual Behavior

`RowParser[Boolean]` throws the following:

> anorm.AnormException: TypeDoesNotMatch(Cannot convert 1: java.lang.Byte to Boolean for column ColumnName(my_table.myBoolean,Some(myBoolean)))

bursauxa:
Fixes #432

### DIFF
--- a/core/src/main/scala/anorm/Column.scala
+++ b/core/src/main/scala/anorm/Column.scala
@@ -280,6 +280,19 @@ object Column extends JodaColumn with JavaTimeColumn {
     }
   }
 
+  private[anorm] val columnByteToBoolean: Column[Boolean] = nonNull { (value, meta) =>
+    val MetaDataItem(qualified, _, _) = meta
+    value match {
+      case bit: Byte if bit == 1 => Right(true)
+      case bit: Byte if bit == 0 => Right(false)
+      case _: Byte => Left(TypeDoesNotMatch(s"Cannot convert Byte $value to Boolean for column $qualified"))
+      case bit: Short if bit == 1 => Right(true)
+      case bit: Short if bit == 0 => Right(false)
+      case _: Short => Left(TypeDoesNotMatch(s"Cannot convert Short $value to Boolean for column $qualified"))
+      case _ => columnToBoolean(value, meta)
+    }
+  }
+
   @SuppressWarnings(Array("AsInstanceOf"))
   private[anorm] def timestamp[T](ts: Timestamp)(f: Timestamp => T): Either[SqlRequestError, T] = Right(if (ts == null) null.asInstanceOf[T] else f(ts))
 
--- a/core/src/main/scala/anorm/package.scala
+++ b/core/src/main/scala/anorm/package.scala
@@ -189,5 +189,22 @@ package object anorm {
   }
 
   /** Activable features */
-  object features {}
+  object features {
+    /**
+     * Column conversion that will accept `Byte` and `Short` values to represent booleans.
+     * This is useful if the underlying database or driver does not support boolean datatype.
+     * For example, the MariaDB JDBC driver v3 will return TINYINT metadata even for columns declared as BOOLEAN.
+     *
+     * 0 is false, 1 is true, anything else is an error.
+     *
+     * Note that the column is not limited to byte or short, boolean types are transparently accepted as well.
+     *
+     * {{{
+     * import anorm.features.columnByteToBoolean
+     * }}}
+     *
+     */
+
+    implicit val columnByteToBoolean: Column[Boolean] = Column.columnByteToBoolean
+  }
 }
--- a/Migration26.md
+++ b/Migration26.md
@@ -197,3 +197,14 @@ The deprecated operations `.list()`, `.single()` and `.singleOpt()` on SQL resul
 The `.getFilledStatement` has been removed.
 
 The former streaming operation `.apply()` is now removed, and must be replaced by either `.fold`, `.foldWhile` or `.withResult`.
+
+## Type mappings
+
+A new column converter is available:
+
+```scala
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -2,3 +2,9 @@ target
 logs
 project/project
 .bsp
+
+# VS Code
```
cchantep, line 6: Please do not IDE specific there, use `.git/info/exclude`

core/src/main/scala/anorm/Column.scala
```diff
@@ -276,6 +276,12 @@ object Column extends JodaColumn with JavaTimeColumn {
     val MetaDataItem(qualified, _, _) = meta
     value match {
       case bool: Boolean => Right(bool)
+      case bit: Byte if bit == 1 => Right(true)
```
cchantep, line 279: As said on the issue, I'm not for having that MariaDB specific behaviour in the common code base.

cchantep (changes requested):

core/src/main/scala/anorm/Column.scala
bursauxa: Well you are the maintainer here, so if you are certain of the decision, then this pull request and the associated issue can be closed.

cchantep: Rather a separate non-default `Column` instance, in `features`, with explicit documentation about MariaDB usage

bursauxa: I had a bit of trouble finding what `features` you were talking about, as it is currently an empty object. I dug up the 2.4 release tag to have an example. Hopefully they way I did it in latest version is what you intended.

core/src/main/scala/anorm/Column.scala
```diff
@@ -280,6 +280,19 @@ object Column extends JodaColumn with JavaTimeColumn {
     }
   }
 
+  private[anorm] val columnToBooleanAcceptingByte: Column[Boolean] = nonNull { (value, meta) =>
```
cchantep, line 283:
```suggestion
  private[anorm] val columnByteToBoolean: Column[Boolean] = nonNull { (value, meta) =>
```

core/src/main/scala/anorm/package.scala
```diff
@@ -189,5 +189,14 @@ package object anorm {
...
-  object features {}
+  object features {
+    /**
+     * Column conversion that will accept `Byte` and `Short` values to represent booleans.
+     * This is useful if the underlying database or driver does not support boolean datatype.
+     * For example, the MariaDB JDBC driver v3 will return TINYINT metadata even for columns declared as BOOLEAN.
+     *
+     * 0 is false, 1 is true, anything else is an error.
+     */
+    implicit val columnToBooleanAcceptingByte: Column[Boolean] = Column.columnToBooleanAcceptingByte
```
cchantep, line 200:
```suggestion
    implicit val columnByteToBoolean: Column[Boolean] = Column.columnToBooleanAcceptingByte
```

core/src/main/scala/anorm/package.scala
```diff
@@ -189,5 +189,14 @@ package object anorm {
...
 
   /** Activable features */
-  object features {}
+  object features {
+    /**
+     * Column conversion that will accept `Byte` and `Short` values to represent booleans.
+     * This is useful if the underlying database or driver does not support boolean datatype.
+     * For example, the MariaDB JDBC driver v3 will return TINYINT metadata even for columns declared as BOOLEAN.
+     *
+     * 0 is false, 1 is true, anything else is an error.
```
cchantep, line 198:
```suggestion
     * 0 is false, 1 is true, anything else is an error.
     *
     * {{{
     * import anorm.features.columnByteToBoolean
     * }}}
```

cchantep (changes requested):

core/src/main/scala/anorm/Column.scala
bursauxa: I performed the renaming as requested, even though I find it slightly misleading (the conversion works for boolean datatype as well, it is not limited to byte or short). I added a line of documentation to clarify this.

cchantep (changes requested):

.gitignore
bursauxa: Sorry I had that changed, but it got lost in rebases

cchantep (approved):

cchantep:
Thx @bursauxa

PR #445, opened 2022-04-10
scala-parser-combinators 2.x only for Scala 3

mkurz:
To avoid any conflicts later on. See:
* https://github.com/playframework/playframework/blob/87d913964d6e809459c6db2e336d092b89b2f955/project/Dependencies.scala#L93-L99
* https://github.com/playframework/twirl/blob/97a60222567d7e2ac5c69830ca86b490616f762f/build.sbt#L15-L20
* https://github.com/playframework/cachecontrol/blob/main/project/Dependencies.scala#L15-L21

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -96,7 +96,7 @@ val armShading = Seq(
 )
 
 lazy val parserCombinatorsVer = Def.setting[String] {
-  if (scalaBinaryVersion.value == "2.11") {
+  if (scalaBinaryVersion.value startsWith "2") {
     "1.1.2"
   } else {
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -2,3 +2,4 @@ target
 logs
 project/project
 .bsp
+.idea/
```
cchantep, line 5: Would rather have it in `.git/info/exclude` as ide specific

mkurz: Removed.

cchantep (approved):

PR #454, opened 2022-05-06
Replace scalariform with scalafmt

mkurz:
sbt-scalariform is not maintained anymore since 2019...

I started migrate this repo to GHA, usually we use `scalafmt[Sbt][Check]All`to check formatting.
scalariformat does not have such a check task, only the format task which actually formats files...


mkurz:
@cchantep I hope you don't mind, I will merge this. It's just formatting anyway.

PR #459, opened 2022-05-15
Introduce sbt-ci-release

mkurz:


mkurz:
@cchantep To release a new version you new just have to locally `git tag <version>` and push that tag, everything else should then be handled by GitHub actions. 
BTW: snapshots will be released for each push to the main branch, see here for example: https://github.com/playframework/anorm/runs/6443488456?check_suite_focus=true#step:5:67

PR #461, opened 2022-05-17
GitHub actions last round + badges

mkurz:


cchantep:
Seems it doesn't work

mkurz:
> Seems it doesn't work: https://github.com/playframework/anorm/runs/7265255275?check_suite_focus=true#step:5:375

@cchantep Looks like this was a temporary problem with sonatype, all other publish jobs aftewards worked, also see https://oss.sonatype.org/content/repositories/snapshots/org/playframework/anorm/anorm-akka_2.13/

cchantep:
I confirm it doesn't work (except for the mentioned snapshot publication)

PR #479, opened 2022-08-16
Fix SQL corresponding type for Java long

### LINKED ISSUE #478: Invalid SQL type computed for Java array prepared statement parameters
### Anorm Version

2.6.x

Mac OS 12.5

### JDK

```
openjdk version "11.0.11" 2021-04-20
OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK-11.0.11+9 (build 11.0.11+9, mixed mode)
```

### Library Dependencies

* Server: PostgreSQL 11.13
* Driver: `"org.postgresql" % "postgresql" % "42.4.1"`

### Expected Behavior

When Java arrays are used as prepared statement parameters:

```scala
        val operatorIds = List(1L, 2L, 3L)
        val operatorIdArray = operatorIds.toArray.map(long2Long)

        SQL"""
        SELECT o.*
        FROM "operators" AS o
        WHERE "operatorId" = any($operatorIdArray)
        """
```

The computed SQL type of array elements should be a `BIGINT` (like in [PostgreSQL documentation](https://www.postgresql.org/docs/current/datatype-numeric.html)).

### Actual Behavior

The computed SQL type of array elements is `NUMERIC`.

pgentile:
# Pull Request Checklist

* [X] Have you added copyright headers to new files?
* [ ] Have you added tests for any changed functionality?

## Fixes

Fixes #478

## Purpose

Fixes `ParameterMetaData` for `Long` type. `Long` should be represented by the SQL `BIGINT` type.

## Background Context

I saw this behavior because it broke an optimized query.

## References

None

### DIFF
--- a/core/src/main/scala/anorm/ParameterMetaData.scala
+++ b/core/src/main/scala/anorm/ParameterMetaData.scala
@@ -130,13 +130,13 @@ object ParameterMetaData extends JavaTimeParameterMetaData {
   }
 
   implicit object LongParameterMetaData extends ParameterMetaData[Long] {
-    val sqlType  = BigIntParameterMetaData.sqlType
-    val jdbcType = BigIntParameterMetaData.jdbcType
+    val sqlType  = "BIGINT"
+    val jdbcType = Types.BIGINT
   }
 
   implicit object JLongParameterMetaData extends ParameterMetaData[JLong] {
-    val sqlType  = BigIntParameterMetaData.sqlType
-    val jdbcType = BigIntParameterMetaData.jdbcType
+    val sqlType  = LongParameterMetaData.sqlType
+    val jdbcType = LongParameterMetaData.jdbcType
   }
 
   /** Decimal (big decimal) parameter meta data */


core/src/main/scala/anorm/ParameterMetaData.scala
```diff
@@ -121,7 +121,7 @@ object ParameterMetaData extends JavaTimeParameterMetaData {
   /** Numeric (big integer) parameter meta data */
   implicit object BigIntParameterMetaData extends ParameterMetaData[BigInt] {
     val sqlType  = "NUMERIC"
-    val jdbcType = Types.BIGINT
+    val jdbcType = Types.NUMERIC
```
cchantep, line 124: That's specific to Pg, so must not be defined there

cchantep: https://github.com/playframework/anorm/tree/main/postgres/src/main/scala

cchantep (changes requested):

core/src/main/scala/anorm/ParameterMetaData.scala
pgentile: I've removed the change. The problem I have noticed occurs only with `long` types.

core/src/main/scala/anorm/ParameterMetaData.scala
```diff
@@ -130,13 +130,13 @@ object ParameterMetaData extends JavaTimeParameterMetaData {
   }
 
   implicit object LongParameterMetaData extends ParameterMetaData[Long] {
-    val sqlType  = BigIntParameterMetaData.sqlType
-    val jdbcType = BigIntParameterMetaData.jdbcType
+    val sqlType  = "BIGINT"
```
cchantep, line 133: Same, specific to PG, so should not happen there

cchantep (changes requested):

PR #591, opened 2023-11-09
Fix revision numbers to ignore reformat

ihostage:
We should be more attention when merge by rebase/squash a scalafmt PRs.


mkurz:
yes... that's because this repo uses rebase by default..

PR #592, opened 2023-11-09
Set Java version to publish explicitly

ihostage:
I think we should set Java version to publish explicitly until we don't decided finally about dropping Java 8 for Anorm (#488).

PS: I know about `-target:jvm-1.8` in `scalacOptions` but anyway.

As a bonus, I added a Mergify config 😄

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,1 @@
+extends: .github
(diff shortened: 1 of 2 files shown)


mkurz:
:+1: I saw that PR and was waiting for CI to become green before merging :wink:

PR #631, opened 2024-12-28
Fix `Dependency Graph` action

ihostage:


ihostage:
Fixed ✅ https://github.com/playframework/anorm/actions/runs/12526506490

PR #697, opened 2026-04-13
Only token changed in release drafter now

mkurz:
I want to quickly try if it is just enough when only the token changed.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: release-drafter/release-drafter@v7
         with:
           name: "Anorm $RESOLVED_VERSION"
-          config-name: .github:release-drafts/increasing-minor-version.yml # located in .github/ in the default branch of the .github repo
+          config-name: release-drafts/increasing-minor-version.yml # located in .github/ in the default branch within this or the .github repo
           commitish: ${{ github.ref_name }}
         env:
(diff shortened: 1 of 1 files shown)


mkurz:
@gaeljw Look at this, all we need(ed) to change is the token, nothing more and everything just works:
```diff
$ git diff 2.11.0..main .github/workflows/release-drafter.yml
diff --git a/.github/workflows/release-drafter.yml b/.github/workflows/release-drafter.yml
index 2aec089..0c933a8 100644
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -9,10 +9,10 @@ jobs:
   update_release_draft:
     runs-on: ubuntu-latest
     steps:
-      - uses: release-drafter/release-drafter@v6
+      - uses: release-drafter/release-drafter@v7
         with:
           name: "Anorm $RESOLVED_VERSION"
           config-name: release-drafts/increasing-minor-version.yml # located in .github/ in the default branch within this or the .github repo
           commitish: ${{ github.ref_name }}
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```

gaeljw:
Correct yes. I thought it's more explicit to add the `.github:` prefix as I initially lost a few minutes before understanding it was not necessarily a reference to the local repo.

But I guess this is common knowledge when using release-drafter.

PR #708, opened 2026-05-15
🧹 chore: baseline Java 17

gaeljw:
# Pull Request Checklist

* [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
* [x] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [x] Have you [squashed your commits](https://www.playframework.com/documentation/2.4.x/WorkingWithGit#Squashing-commits)?
* [ ] Have you added copyright headers to new files?
* [ ] Have you checked that both Scala and Java APIs are updated?
* [ ] Have you updated the documentation for both Scala and Java sections?
* [ ] Have you added tests for any changed functionality?

## Fixes

Solves #707 ; built on top of #709 (read only the 2nd commit)

## Purpose

Generate Java 17 class files. Upgrade the minimum required Java version to 17 for users.

PR #709, opened 2026-05-15
🧹 chore: drop support for Scala 2.12 (#689)

gaeljw:
# Pull Request Checklist

* [ ] Have you added copyright headers to new files?
* [ ] Have you added tests for any changed functionality?

# Fixes

Closes #689 

# Additional details

As part of this, the module `anorm-iteratee` is removed as it was only published for Scala 2.12.


.github/workflows/build-test.yml
```diff
@@ -23,34 +23,14 @@ jobs:
     name: Binary Compatibility
     uses: playframework/.github/.github/workflows/binary-check.yml@v4
 
-  check-docs-8:
```
gaeljw, line 26: ℹ️ I removed it because the combo Java 8 and Scala 2.13 doesn't work

```
Exception in thread "specs2-3" java.lang.BootstrapMethodError: java.lang.UnsupportedClassVersionError: play/api/test/WithApplication has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

This because our setup for Scala 2.13 pulls Play 3.x which requires Java 11+.

As we're gonna baseline on Java 17 soon ( #708 ), I don't feel it's worth making it work by changing the Play version targeted.

Actually I'm not even sure what was the purpose of this double pipeline in Java 8 & 11 but not testing all Scala version for Java 8. Anyway, as I said, we're dropping Java 8.

PR #712, opened 2026-05-19
🐛 fix: surface UnexpectedNullableFound instead of remapping to ColumnNotFound (closes #560)

### LINKED ISSUE #560: Misleading error message when column is unexpectedly NULL
### Are you looking for help?

No

### Anorm Version (2.5.x / etc)

2.7.0 (latest version)

Use `uname -a` if on Linux.

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

Adopt JDK 11

### Library Dependencies

N/A

### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

1.  Use the `scalar[Int].single` on a column that unexpectedly contains a NULL
2. An error is thrown that says the column unexpectedly contains a null

### Actual Behavior

1.  Use the `scalar[Int].single` on a column that unexpectedly contains a NULL
2. This error is thrown: `anorm.AnormException: '.completed_average' not found, available columns: completed_average, completed_average`
3. The error makes it look like there was an issue with column name or column alias etc.  The actual issue was that the column unexpectedly contains a NULL and the `scalar[Option[Int]].single` needs to be used instead

The issue seems to be with this code in Anorm: 

```scala
  @inline private def parseColumn[T](
      row: Row,
      name: String,
      c: Column[T],
      input: (Any, MetaDataItem)
  ): Either[SqlRequestError, T] = c.tupled(input).left.map {
    case UnexpectedNullableFound(_) =>
      ColumnNotFound(name, row)

    case cause => cause
  }
```

The `UnexpectedNullableFound` error is incorrectly mapped to the `ColumnNotFound` error

### Reproducible Test Case

Please provide a PR with a failing test.  

If the issue is more complex or requires configuration, please provide a link to a project on Github that reproduces the issue.

MavenRain:
Closes #560.

  ## Problem

  `scalar[T].single` (and other non-Option parsers) on a column containing
  NULL produced this confusing error:

  '.completed_average' not found, available columns: completed_average,
  completed_average

  The column was actually present; it just contained NULL.  The root cause
  was an internal `parseColumn` helper in `SqlParser` that remapped
  `UnexpectedNullableFound` to `ColumnNotFound`.  @gaeljw confirmed the
  impact in
  https://github.com/playframework/anorm/issues/560#issuecomment-1911862878.

  ## Fix

  Two parts:

  1. **`SqlParser.scala`:** drop the `UnexpectedNullableFound → ColumnNotFound`
     remap in `parseColumn`.  Since the helper then only forwarded to
     `c.tupled(input)`, it has been inlined at its three call sites and
     removed.  Errors from `Column[T].tupled` now surface unchanged, so
     users see `UnexpectedNullableFound(column_name)` and can distinguish
     NULL from missing column.

  2. **`RowParser.scala` (`.?` combinator):** the optional-column combinator
     previously only recovered from `ColumnNotFound`.  Its scaladoc has
     always promised "missing or null column as None", and the existing
     `singleOpt` / `?` semantics depended on the buggy mapping to convert
     NULL to `None`.  Updated `.?` to also recover from
     `UnexpectedNullableFound`, honouring its documented contract
     independently of the `parseColumn` remap.

  ## Regression test

  `AnormSpec` gains one test verifying that `scalar[String].single` on a
  NULL result throws an error whose message contains
  `UnexpectedNullableFound` and **does not** contain
  `not found, available columns`.

  ## Behaviour change

  User code that catches `ColumnNotFound` specifically to recognise a NULL
  column will now see `UnexpectedNullableFound` instead.  This was the
  documented behaviour all along; the buggy remap was the deviation.  No
  other observable change.

  ## Local validation

  - `++ 2.13; anorm-core/compile anorm-core/Test/compile` clean
  - `++ 3;    anorm-core/compile anorm-core/Test/compile` clean
  - `++ 2.13; anorm-core/test` 294 / 294 pass (including the new test and
    the existing "be parsed from mapping with optional column" test which
    previously depended on the buggy mapping)
  - `++ 3; anorm-core/testOnly tests.AnormSpec` 52 / 52 pass
  - `scalafmtCheckAll` clean
  - `anorm-core/mimaReportBinaryIssues` no binary issues

### DIFF
--- a/core/src/main/scala/anorm/RowParser.scala
+++ b/core/src/main/scala/anorm/RowParser.scala
@@ -97,10 +97,16 @@ trait RowParser[+A] extends (Row => SqlResult[A]) { parent =>
    * that will turn missing or null column as None.
    */
   def ? : RowParser[Option[A]] = RowParser {
+    // Note: `UnexpectedNullableFound` is the error raised by `Column` when a
+    // non-nullable column is read but contains NULL.  Previously the SqlParser
+    // remapped this to `ColumnNotFound` so that this combinator surfaced NULL
+    // as `None`; that remap also caused #560 (misleading "column not found"
+    // errors elsewhere).  Now we recover from both errors here directly, which
+    // is what the scaladoc above already promised.
     parent(_) match {
-      case Success(a)                  => Success(Some(a))
-      case Error(ColumnNotFound(_, _)) =>
-        Success(None)
+      case Success(a)                        => Success(Some(a))
+      case Error(ColumnNotFound(_, _))       => Success(None)
+      case Error(UnexpectedNullableFound(_)) => Success(None)
 
       case e @ Error(_) => e
     }
--- a/core/src/main/scala/anorm/SqlParser.scala
+++ b/core/src/main/scala/anorm/SqlParser.scala
@@ -29,8 +29,12 @@ object SqlParser extends FunctionAdapter with DeprecatedSqlParser {
         } yield v -> m).toRight(NoColumnsInReturnedResult)
 
         val parsed = Compat.rightFlatMap[SqlMappingError, SqlRequestError, (Any, MetaDataItem), T](input) {
-          case in @ (_, m) =>
-            parseColumn(row, m.column.qualified, c, in)
+          case in @ (_, _) =>
+            // Previously a `parseColumn` helper mapped `UnexpectedNullableFound` to
+            // `ColumnNotFound` here, producing misleading "column not found" errors
+            // when the column existed but contained NULL (#560).  Surface the
+            // original `Column` error instead.
+            c.tupled(in)
         }
 
         parsed.fold(Error(_), Success(_))
@@ -491,7 +495,7 @@ object SqlParser extends FunctionAdapter with DeprecatedSqlParser {
     row =>
       Compat
         .rightFlatMap(row.get(name)) { in =>
-          parseColumn(row, name, c, in)
+          c.tupled(in)
         }
         .fold(Error(_), Success(_))
   }
@@ -514,7 +518,7 @@ object SqlParser extends FunctionAdapter with DeprecatedSqlParser {
     RowParser { row =>
       Compat
         .rightFlatMap(row.getIndexed(position - 1)) { in =>
-          parseColumn(row, in._2.column.qualified, c, in)
+          c.tupled(in)
         }
         .fold(Error(_), Success(_))
     }
@@ -536,17 +540,6 @@ object SqlParser extends FunctionAdapter with DeprecatedSqlParser {
   def matches[T: Column](column: String, value: T): RowParser[Boolean] =
     get[T](column).?.map(_.fold(false) { _ == value })
 
-  @inline private def parseColumn[T](
-      row: Row,
-      name: String,
-      c: Column[T],
-      input: (Any, MetaDataItem)
-  ): Either[SqlRequestError, T] = c.tupled(input).left.map {
-    case UnexpectedNullableFound(_) =>
-      ColumnNotFound(name, row)
-
-    case cause => cause
-  }
 }
 
 @deprecated("Do not use these combinators", "2.5.4")
--- a/core/src/test/scala/AnormSpec.scala
+++ b/core/src/test/scala/AnormSpec.scala
@@ -66,6 +66,21 @@ final class AnormSpec
 
       }
 
+      // Regression test for #560: previously this surfaced as
+      // "'X' not found, available columns: X, X" because parseColumn mapped
+      // UnexpectedNullableFound to ColumnNotFound.  After the fix it must
+      // mention the column is NULL, not missing.
+      "surface NULL-aware error when scalar[T].single hits NULL" in withQueryResult(null.asInstanceOf[String]) {
+        implicit c: Connection =>
+
+          SQL("SELECT * FROM test").as(scalar[String].single).aka("scalar single on NULL") must throwA[Exception]
+            .like {
+              case e: Exception =>
+                (e.getMessage.aka("error") must contain("UnexpectedNullableFound"))
+                  .and(e.getMessage.aka("error") must not(contain("not found, available columns")))
+            }
+      }
+
       "throw exception when single result is missing" in withQueryResult(fooBarTable) { implicit c: Connection =>
 
         SQL("SELECT * FROM test").as(fooBarParser1.single).aka("mapping") must throwA[Exception].like {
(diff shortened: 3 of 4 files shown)


core/src/main/scala/anorm/RowParser.scala
```diff
@@ -97,10 +97,16 @@ trait RowParser[+A] extends (Row => SqlResult[A]) { parent =>
...
+    // as `None`; that remap also caused #560 (misleading "column not found"
+    // errors elsewhere).  Now we recover from both errors here directly, which
+    // is what the scaladoc above already promised.
     parent(_) match {
-      case Success(a)                  => Success(Some(a))
-      case Error(ColumnNotFound(_, _)) =>
-        Success(None)
+      case Success(a)                        => Success(Some(a))
+      case Error(ColumnNotFound(_, _))       => Success(None)
+      case Error(UnexpectedNullableFound(_)) => Success(None)
```
cchantep, line 109: If that's the desired behaviour, the pattern could be merge with the previous case

core/src/main/scala/anorm/RowParser.scala
```diff
@@ -97,10 +97,16 @@ trait RowParser[+A] extends (Row => SqlResult[A]) { parent =>
    * that will turn missing or null column as None.
    */
   def ? : RowParser[Option[A]] = RowParser {
+    // Note: `UnexpectedNullableFound` is the error raised by `Column` when a
```
gaeljw, line 100: I don't think we need this entire comment. Sounds like classic AI over commenting ;)

core/src/main/scala/anorm/SqlParser.scala
```diff
@@ -29,8 +29,12 @@ object SqlParser extends FunctionAdapter with DeprecatedSqlParser {
         } yield v -> m).toRight(NoColumnsInReturnedResult)
 
         val parsed = Compat.rightFlatMap[SqlMappingError, SqlRequestError, (Any, MetaDataItem), T](input) {
-          case in @ (_, m) =>
-            parseColumn(row, m.column.qualified, c, in)
+          case in @ (_, _) =>
+            // Previously a `parseColumn` helper mapped `UnexpectedNullableFound` to
```
gaeljw, line 33: Same comment, I don't think the comment is useful.

gaeljw:
Thanks for the PR @MavenRain , this is pretty much the same as what Copilot already suggested in https://github.com/playframework/anorm/pull/665 but I haven't had time to dig a bit more at the consequences of the changes as I'm not yet familiar with the codebase and was a bit reluctant to merge without better understanding.

I'll try to find some time soon to review this.

Unless this is straightforward good for you @cchantep :)

cchantep:
> Thanks for the PR @MavenRain , this is pretty much the same as what Copilot already suggested in #665 but I haven't had time to dig a bit more at the consequences of the changes as I'm not yet familiar with the codebase and was a bit reluctant to merge without better understanding.
> 
> I'll try to find some time soon to review this.
> 
> Unless this is straightforward good for you @cchantep :)

Not straightforward, needs cautious review.
If there is already an ongoing PR, would rather go on with.

MavenRain:
Thanks @gaeljw and @cchantep.  Agreed that #665 came first;  I'm happy to close in its favour if that's the call.  Two notes before deciding:

  1. **#665 has been dormant since 2026-02-28** (still draft, authored by
     `copilot-swe-agent`), so review iteration there is one-way: any
     change-request would need a maintainer to push commits onto the bot's
     branch or re-trigger it.

  2. **#712 has one small cleanup #665 lacks.**  Once the
     `UnexpectedNullableFound → ColumnNotFound` remap is removed,
     `SqlParser.parseColumn` just forwards to `c.tupled(input)`.  #712
     inlines it at its three call sites and deletes the helper; #665
     leaves the vestigial helper in place.

I've pushed an update that rebases #712 onto current `main` and, for parity with #665, ports its `SqlParser.int("n").?` NULL test into `SqlResultSpec` (alongside the existing `str("foo").?` coverage on the `Column[String]` codepath).  Local validation on the new base:

  - `++ 2.13; anorm-core/Test/compile` and `++ 3; anorm-core/Test/compile`:
  clean
  - `++ 2.13; anorm-core/testOnly anorm.SqlResultSpec tests.AnormSpec`: 89/89
  - `++ 3;    anorm-core/testOnly anorm.SqlResultSpec tests.AnormSpec`: 88/88
  - `scalafmtCheckAll`: clean
  - `anorm-core/mimaReportBinaryIssues`: no issues

If you'd rather revive #665, then I'm happy to close this;  I'm just flagging that path needs a human re-author for any review iteration.

MavenRain:
@gaeljw @cchantep: friendly bump 🙂 Still green and rebased on main.  On the caller-impact question: removing the remap doesn't change the `.?` behavior (it now recovers from `UnexpectedNullableFound` directly: see the `new int("n").?.single` must beNone test), so the only visible change is a non-optional NULL read reporting the accurate `UnexpectedNullableFound` instead of the misleading `ColumnNotFound` from #560.  I'm happy to add a migration note, or to defer to #665 if you'd prefer, although that one's been a dormant bot draft since Feb.  Whatever's least work for you both is fine.

gaeljw:
I know this is taking more time than you'd expect but as I mentioned earlier this is a piece of code that I haven't looked at yet and I definitely want to have good understanding of it before including a change that will affect users.

I'll try to free some time in the coming days.

Did you test the change on some projects of yours that use Anorm and for which you faced the exception?

MavenRain:
> I know this is taking more time than you'd expect but as I mentioned earlier this is a piece of code that I haven't looked at yet and I definitely want to have good understanding of it before including a change that will affect users.
> 
> I'll try to free some time in the coming days.
> 
> Did you test the change on some projects of yours that use Anorm and for which you faced the exception?

Honest answer: No, I didn't hit this in a project of my own;  I came to it from the issue rather than my own stack trace.  So instead of a production app, I built a minimal standalone repro that runs against a real in-memory H2 database (real JDBC `NULL`, not the acolyte harness): https://github.com/MavenRain/anorm-560-repro
  
It runs three scenarios against whichever anorm you point it at (default = released 3.0.0;  set `ANORM_VERSION` to a local publish of this branch).
 
Verified results:
```  
   scalar[Int].single on a present-but-NULL column
       released 3.0.0 : AnormException: '...COMPLETED_AVERAGE' not found,
   available columns: ...   (the #560 mislabel)
       this branch    : AnormException: UnexpectedNullableFound(ColumnName(...))
                     (accurate)
  
   int("...").?.single on the same NULL column
       released 3.0.0 : None
       this branch    : None        (optional path unchanged)
  
   normal non-NULL read
       released 3.0.0 : 42
       this branch    : 42
```  

So the only behavioural change is the non-optional read going from the misleading "not found" to the accurate `UnexpectedNullableFound`;  the optional `.?` path stays `None` either way.  The README has copy-paste steps to publish the branch locally and reproduce both columns.  I'd be happy to also add a CHANGELOG/migration note for anyone matching on `ColumnNotFound`.   No rush . . . take the time you need.

gaeljw:
I finally merged #665 as the changes were identical to this one in the end and I tested it on a real project.

Thanks for pushing about this topic @MavenRain

PR #714, opened 2026-05-21
🐛 fix: preserve nanosecond precision in Column[LocalDateTime] (#525)

### LINKED ISSUE #525: Default Column[LocalDateTime] instance lose nanoseconds precision
I'll submit this in the same tradition as #408  and #325 .

I updated the repro from #408 to `LocalDateTime` to show the same problem again.

https://github.com/oyvindberg/play-anorm-nano-precision
```
[info] - should insert a LocalDateTime WITHOUT nano precision and return the inserted value
[info] - should insert a LocalDateTime instant WITH nano precision and return the inserted value *** FAILED ***
[info]   2021-01-06T09:22:31.222 was not equal to 2021-01-06T09:22:31.222456 (ZonedDateTimeRepositoryTest.scala:62)
```

MavenRain:
Closes #525.

  ## Problem

  `scalar[LocalDateTime].single` on a `Timestamp` column with sub-millisecond
  precision truncates the nanoseconds (reproducer in
  https://github.com/oyvindberg/play-anorm-nano-precision):

  [info] - should insert a LocalDateTime instant WITH nano precision and return
  the inserted value *** FAILED ***
  [info]   2021-01-06T09:22:31.222 was not equal to 2021-01-06T09:22:31.222456

  Same bug class as #325 (`Instant`) and #408 (`ZonedDateTime`), fixed by #326
  and #409 respectively.

  ## Fix

  Route `columnToLocalDateTime`'s non-`LocalDateTime` cases through
  `instantValueTo` instead of `temporalValueTo`.  `instantValueTo` uses
  `Timestamp#toInstant` (which preserves nanos), while `temporalValueTo` went
  through `Timestamp#getTime → Instant.ofEpochMilli` (millis only).

  ## Regression test

  `JavaTimeColumnSpec` gains a "be parsed from timestamp with nano precision"
  case parallel to the existing `Instant` and `ZonedDateTime` versions.

  ## Local validation

  - `++ 2.13; anorm-core/compile anorm-core/Test/compile` clean
  - `++ 3; anorm-core/compile anorm-core/Test/compile` clean
  - `++ 2.13; anorm-core/test` 294 / 294 pass (including the new nano test)
  - `++ 3; anorm-core/testOnly anorm.JavaTimeColumnSpec` 27 / 27 pass
  - `scalafmtCheckAll` clean
  - `anorm-core/mimaReportBinaryIssues` no binary issues

### DIFF
--- a/core/src/main/scala/anorm/Column.scala
+++ b/core/src/main/scala/anorm/Column.scala
@@ -720,15 +720,19 @@ sealed trait JavaTimeColumn {
    * }}}
    */
   implicit val columnToLocalDateTime: Column[LocalDateTime] = {
-    def millisToLocalDateTime(ts: Long) =
-      LocalDateTime.ofInstant(Instant.ofEpochMilli(ts), ZoneId.systemDefault)
+    def instantToLocalDateTime(instant: Instant) =
+      LocalDateTime.ofInstant(instant, ZoneId.systemDefault)
 
     nonNull { (value, meta) =>
       value match {
         case localDateTime: LocalDateTime => Right(localDateTime)
 
         case _ =>
-          temporalValueTo[LocalDateTime](millisToLocalDateTime, "Java8 LocalDateTime")(value, meta)
+          // Route through instantValueTo so Timestamp values keep nanosecond
+          // precision (via Timestamp#toInstant) instead of being truncated to
+          // milliseconds through Timestamp#getTime.  See #525, mirrors the
+          // fixes in #326 (Instant) and #409 (ZonedDateTime).
+          instantValueTo[LocalDateTime](instantToLocalDateTime, "Java8 LocalDateTime")(value, meta)
       }
     }
   }
--- a/core/src/main/scala/anorm/Column.scala
+++ b/core/src/main/scala/anorm/Column.scala
@@ -719,21 +719,15 @@ sealed trait JavaTimeColumn {
    *     as(SqlParser.scalar[LocalDateTime].single)
    * }}}
    */
-  implicit val columnToLocalDateTime: Column[LocalDateTime] = {
-    def instantToLocalDateTime(instant: Instant) =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/anorm/Column.scala
```diff
@@ -720,15 +720,19 @@ sealed trait JavaTimeColumn {
...
+    def instantToLocalDateTime(instant: Instant) =
+      LocalDateTime.ofInstant(instant, ZoneId.systemDefault)
 
     nonNull { (value, meta) =>
       value match {
         case localDateTime: LocalDateTime => Right(localDateTime)
 
         case _ =>
-          temporalValueTo[LocalDateTime](millisToLocalDateTime, "Java8 LocalDateTime")(value, meta)
+          // Route through instantValueTo so Timestamp values keep nanosecond
```
gaeljw: Please remove these comments, they are useful in the review context but not for the long term codebase maintenance. The test is enough to cover the intended behaviour.

core/src/main/scala/anorm/Column.scala
```diff
@@ -720,15 +720,19 @@ sealed trait JavaTimeColumn {
    * }}}
    */
   implicit val columnToLocalDateTime: Column[LocalDateTime] = {
-    def millisToLocalDateTime(ts: Long) =
-      LocalDateTime.ofInstant(Instant.ofEpochMilli(ts), ZoneId.systemDefault)
+    def instantToLocalDateTime(instant: Instant) =
+      LocalDateTime.ofInstant(instant, ZoneId.systemDefault)
```
gaeljw: You can remove this intermediate method and inline it directly in `instantValueTo` like it's done in other methods. The intermediate method is not useful anymore in terms of readability IMHO.

gaeljw:
Thanks @MavenRain , two minor comments and I'll merge it :)

gaeljw (approved):

